PR #5, opened 2022-12-22
Feature/move hangfire to quartz

RodriFS:
The retry count for the spontaneous jobs for opening of channels, closing of channels and withdrawals is 20 retries, and 10 seconds of delay between each retry for now. We can put it in an env variable if needed.

Quartz migration uses a sql file you can find here: https://github.com/quartznet/quartznet/blob/main/database/tables/tables_postgres.sql


.vscode/launch.json
```diff
@@ -58,7 +58,7 @@
                 "NBXPLORER_URI": "http://127.0.0.1:32838",
                 "NBXPLORER_BTCRPCUSER": "polaruser",
                 "NBXPLORER_BTCRPCPASSWORD": "polarpass",
-                "NBXPLORER_BTCRPCURL": "http://127.0.0.1:18443/",
```
Jossec101, line 61: I think we are having a clash here @RodriFS, I have to change this all the time

src/Data/Repositories/ChannelRepository.cs
```diff
@@ -112,9 +112,31 @@ public async Task<(bool, string?)> SafeRemove(Channel type, bool forceClose = fa
 
             if (closeRequest == null) return (false, null);
 
-            var jobId = _backgroundJobClient.Enqueue<ILightningService>(service => service.CloseChannel(closeRequest, forceClose));
-
-            closeRequest.JobId = jobId;
+            IScheduler scheduler = await _schedulerFactory.GetScheduler();
```
Jossec101, line 115:
```suggestion
            var scheduler = await _schedulerFactory.GetScheduler(); 
```

src/Helpers/PriceConversionHelper.cs
```diff
@@ -26,7 +26,7 @@ public static decimal GetBtcToUsdPrice()
         try
         {
             JsonDocument document = JsonDocument.Parse(response.Content);
-            btcPrice = document.RootElement[0].GetProperty("current_price").GetDecimal();
+            btcPrice = document.RootElement.GetProperty("current_price").GetDecimal();
```
Jossec101, line 29: What about this change?

src/Pages/Withdrawals.razor
```diff
@@ -652,12 +653,32 @@
...
+                                        .Build();
+                                
+                                var trigger = TriggerBuilder.Create()
+                                    .WithIdentity($"{nameof(PerformWithdrawalJob)}Trigger")
+                                    .StartNow()
+                                    .WithSimpleSchedule(opts =>
+                                        opts
+                                            .WithIntervalInSeconds(10)
+                                            .WithRepeatCount(20))
+                                    .Build();
```
Jossec101, line 674: I see this pattern repeated a lot, maybe it could be refactored in just a method, but its ok for now

Jossec101:
@RodriFS Check the conflicts please

.vscode/launch.json
RodriFS: This change was done because polar's bitcoind defaults to 18445, and we can use the polar UI to fund and create channels with the `docker-compose.dev-novs.yml` file. What docker compose are you using? Maybe we can also change that compose file to use 18445

src/Pages/Withdrawals.razor
Jossec101: I added a new issue for this @RodriFS

src/Helpers/PriceConversionHelper.cs
RodriFS: It was my error. When coingeko fails, instead of throwing, it returns an "Object" with an error message. Now I check if it is an array before getting the price

src/Pages/Withdrawals.razor
RodriFS: Excellent

.vscode/launch.json
RodriFS: Changed back

src/Data/Repositories/ChannelRepository.cs
RodriFS: done

Jossec101 (approved):
LGTM

Issue #4, opened 2022-12-22
Setup in production overhaul


Jossec101:
Right now, in production mode, NodeGuard will generate its internal wallet automatically and print the wallet seedphrase in the console, but there is a security/privacy issue. We will implement a prompt on start to fix this.

When NodeGuard is opened for the first time (i.e. No superadmin users), we allow to set a new superadmin / password, then if not internal wallets are in the db, the same UI will generate one and display it on the screen for the user to store it safely.

Jossec101:
New wiki page https://github.com/Elenpay/NodeGuard/wiki/Initial-setup

PR #7, opened 2022-12-27
Initial setup for production overhaul

Jossec101:
Initial setup for production overhaul #4

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -39,7 +39,7 @@ services:
         COINGECKO_ENDPOINT: https://pro-api.coingecko.com/api/v3/coins/markets?vs_currency=usd&ids=bitcoin
         COINGECKO_KEY: "TBD" #PLEASE FILL IT WITH YOUR OWN
         MONITOR_WITHDRAWALS_CRON: "0 */1 * * * ?" # Every Minute
-        ENABLE_REMOTE_FM_SIGNER: "false" #Changes the way of signing, instead of a auto-generated internal wallet, the FM will sign with a remote signer function on AWS Lambda
+        ENABLE_REMOTE_SIGNER: "false" #Changes the way of signing, instead of a auto-generated internal wallet, the FM will sign with a remote signer function on AWS Lambda
         MSSQL_CONNECTIONSTRING: "Server=localhost;Database=master;User=sa;Password=rw_dev;"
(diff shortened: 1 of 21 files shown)


docker/docker-compose.yml
```diff
@@ -42,7 +39,7 @@ services:
         COINGECKO_ENDPOINT: https://pro-api.coingecko.com/api/v3/coins/markets?vs_currency=usd&ids=bitcoin
         COINGECKO_KEY: "TBD" #PLEASE FILL IT WITH YOUR OWN
         MONITOR_WITHDRAWALS_CRON: "0 */1 * * * ?" # Every Minute
-        ENABLE_REMOTE_FM_SIGNER: "false" #Changes the way of signing, instead of a auto-generated internal wallet, the FM will sign with a remote signer function on AWS Lambda
+        ENABLE_REMOTE_SIGNER: "false" #Changes the way of signing, instead of a auto-generated internal wallet, the FM will sign with a remote signer function on AWS Lambda
```
markettes, line 42: FM still in the comment

src/Migrations/ApplicationDbContextModelSnapshot.cs
```diff
@@ -264,12 +264,14 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                         .HasColumnType("text");
 
                     b.Property<string>("MnemonicString")
-                        .IsRequired()
                         .HasColumnType("text");
 
                     b.Property<DateTimeOffset>("UpdateDatetime")
                         .HasColumnType("timestamp with time zone");
 
+                    b.Property<string>("XPUB")
```
markettes, line 272: Is the XPUB not required?

docker/docker-compose.yml
Jossec101:
```suggestion
        ENABLE_REMOTE_SIGNER: "false" #Changes the way of signing, instead of a auto-generated internal wallet, NodeGuard will sign with a remote signer function on AWS Lambda
```

src/Migrations/ApplicationDbContextModelSnapshot.cs
Jossec101: Yes because it can be null at the creation of the entity.

markettes (approved):
LGTM

RodriFS (approved):

PR #32, opened 2023-01-05
Retriable jobs with configurable intervals

RodriFS:
Added some env variables for configuring intervals of jobs but they default to an array of  [1, 5, 10, 20] minutes.

The first time the job triggers without any delay, and then it starts consuming the array and rescheduling the interval.

For env variables, the format is a string of numbers separated by commas with no spaces. Example:
```
JOB_RETRY_LIST_CHANNEL_OPEN="2,4,8,16,32"
```

### DIFF
--- a/src/Data/Repositories/ChannelRepository.cs
+++ b/src/Data/Repositories/ChannelRepository.cs
@@ -118,7 +118,9 @@ namespace FundsManager.Data.Repositories
             var map = new JobDataMap();
             map.Put("closeRequestId", closeRequest.Id);
             map.Put("forceClose", forceClose);
-            var job = RetriableJob.Create<ChannelCloseJob>(map, closeRequest.Id.ToString());
+
+            var retryList = RetriableJob.ParseRetryListFromEnvironmenVariable("JOB_RETRY_LIST_CHANNEL_CLOSE");
+            var job = RetriableJob.Create<ChannelCloseJob>(map, closeRequest.Id.ToString(), retryList);
             await scheduler.ScheduleJob(job.Job, job.Trigger);
 
             // TODO: Check job id
--- a/src/Data/Repositories/ChannelRepository.cs
+++ b/src/Data/Repositories/ChannelRepository.cs
@@ -119,7 +119,7 @@ namespace FundsManager.Data.Repositories
             map.Put("closeRequestId", closeRequest.Id);
             map.Put("forceClose", forceClose);
 
-            var retryList = RetriableJob.ParseRetryListFromEnvironmenVariable("JOB_RETRY_INTEVAL_LIST_IN_MINUTES");
+            var retryList = RetriableJob.ParseRetryListFromEnvironmenVariable("JOB_RETRY_INTERVAL_LIST_IN_MINUTES");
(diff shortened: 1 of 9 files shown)


src/Jobs/ProcessNodeChannelAcceptorJob.cs
```diff
@@ -227,7 +227,7 @@ await foreach (var response in resultAcceptor.ResponseStream.ReadAllAsync())
         catch (Exception e)
         {
             _logger.LogError(e, "Error on {JobName}", nameof(ProcessNodeChannelAcceptorJob));
-            throw new JobExecutionException(e, true);
+            throw new JobExecutionException(e, false);
```
RodriFS, line 230: I modified this cronjobs to not refire immediately after an error, but to retry on the next interval. Let me know if this is ok.

Jossec101: I would say so, this is what hangfire usually does

src/Data/Repositories/ChannelRepository.cs
```diff
@@ -118,7 +118,9 @@ public async Task<(bool, string?)> SafeRemove(Channel type, bool forceClose = fa
             var map = new JobDataMap();
             map.Put("closeRequestId", closeRequest.Id);
             map.Put("forceClose", forceClose);
-            var job = RetriableJob.Create<ChannelCloseJob>(map, closeRequest.Id.ToString());
+
+            var retryList = RetriableJob.ParseRetryListFromEnvironmenVariable("JOB_RETRY_LIST_CHANNEL_CLOSE");
```
Jossec101, line 122: Why CHANNEL_CLOSE? I mean this retry env var should be for every single retriable job, no matter the context or what it does. Maybe we can use something like JOB_RETRY_INTERVALS, a better name can be found for sure

src/Helpers/JobTypes.cs
```diff
@@ -36,12 +54,57 @@ public class RetriableJob
...
+            .Select<string, int>(s => int.Parse(s))
+            .ToArray();
+    }
+
+    /// <summary>
+    /// Call this function inside your Job's Execute method to set up retries and execute your action.
+    /// </summary>
+    /// <param name="context">The execution context of the job</param>
+    /// <param name="f">The action you want to perform inside the job</param>
+    public static async Task Execute(IJobExecutionContext context, Func<Task> f)
```
Jossec101, line 78: f is a single char for a variable name, I would recommend writing what it does in a c#-ish way, like `function` or `callback`

Jossec101 (commented):
Quick question, so this would trigger as many times as the lenght of the interval env-var right?

src/Data/Repositories/ChannelRepository.cs
RodriFS: sure, I thought we might want to handle them separately but we can do just one env var for all. I'll do the change

> Quick question, so this would trigger as many times as the lenght of the interval env-var right?

Yes, this would trigger as many times as the length of the interval you set up.

src/Helpers/JobTypes.cs
RodriFS: I agree, I'll change it

Jossec101 (approved):
LGTM

PR #43, opened 2023-01-09
38 nodeguard withdrawals only allow to add negative btc amount

markettes:
Done:
- Small fix


src/Helpers/ValidationHelper.cs
```diff
@@ -58,11 +58,14 @@ public static void ValidateWithdrawalAmount(ValidatorEventArgs obj, Boolean isAm
...
-        var environmentVariableMax = Environment.GetEnvironmentVariable("MAXIMUM_WITHDRAWAL_BTC_AMOUNT") ?? 21_000_000m;
-        var minimum = decimal.Parse(environmentVariableMin, NumberStyles.AllowDecimalPoint, CultureInfo.InvariantCulture);
-        var maximum = decimal.Parse(environmentVariableMax, NumberStyles.AllowDecimalPoint, CultureInfo.InvariantCulture);
-        
+        var environmentVariableMin = Environment.GetEnvironmentVariable("MINIMUM_WITHDRAWAL_BTC_AMOUNT");
+        var environmentVariableMax = Environment.GetEnvironmentVariable("MAXIMUM_WITHDRAWAL_BTC_AMOUNT");
+        decimal minimum, maximum;
+        if (environmentVariableMin == null) minimum = 0.0m;
+        else minimum = decimal.Parse(environmentVariableMin, NumberStyles.AllowDecimalPoint, CultureInfo.InvariantCulture);
+        if (environmentVariableMax == null) maximum = 21_000_000;
```
RodriFS, line 66: is the same without the "m"?

Jossec101: Its a way of telling C# compiler to use decimals, but sometimes it is not strictly needed, I did it just in case because I was not using a IDE/Editor

src/Helpers/ValidationHelper.cs
```diff
@@ -58,11 +58,14 @@ public static void ValidateWithdrawalAmount(ValidatorEventArgs obj, Boolean isAm
...
 
-        var environmentVariableMin = Environment.GetEnvironmentVariable("MINIMUM_WITHDRAWAL_BTC_AMOUNT") ?? 0.0m;
-        var environmentVariableMax = Environment.GetEnvironmentVariable("MAXIMUM_WITHDRAWAL_BTC_AMOUNT") ?? 21_000_000m;
-        var minimum = decimal.Parse(environmentVariableMin, NumberStyles.AllowDecimalPoint, CultureInfo.InvariantCulture);
-        var maximum = decimal.Parse(environmentVariableMax, NumberStyles.AllowDecimalPoint, CultureInfo.InvariantCulture);
-        
+        var environmentVariableMin = Environment.GetEnvironmentVariable("MINIMUM_WITHDRAWAL_BTC_AMOUNT");
+        var environmentVariableMax = Environment.GetEnvironmentVariable("MAXIMUM_WITHDRAWAL_BTC_AMOUNT");
+        decimal minimum, maximum;
+        if (environmentVariableMin == null) minimum = 0.0m;
```
Jossec101, line 64:
```suggestion
        if (environmentVariableMin == null) minimum = 0.0;
```

Jossec101 (approved):

PR #46, opened 2023-01-10
38 nodeguard withdrawals only allow to add negative btc amount

markettes:
Done: 
- Applied same logic to the max and min in the numeric field as for validation in save


src/Pages/Withdrawals.razor
```diff
@@ -722,22 +717,22 @@
 
     private async Task RejectOrCancel()
     {
-        if ((_rejectCancelStatus == WalletWithdrawalRequestStatus.Cancelled || _rejectCancelStatus == WalletWithdrawalRequestStatus.Rejected)
```
cvicenti, line 725: did you perform these formatting changes on the go or  were they automatically done by your IDE? We should keep merge requests small and change only what it is described in the issue IMHO.
We can always create a new issue to address formatting issue

src/Pages/Withdrawals.razor
```diff
@@ -440,19 +440,14 @@
...
-        {
-            ToastService.ShowError("Getting maximum of Bitcoins for withdrawal has failed");
-        }
-        
+
+        var environmentVariableMin = Environment.GetEnvironmentVariable("MINIMUM_WITHDRAWAL_BTC_AMOUNT");
+        var environmentVariableMax = Environment.GetEnvironmentVariable("MAXIMUM_WITHDRAWAL_BTC_AMOUNT");
+        if (environmentVariableMin == null) _minWithdrawal = 0.0m;
+        else _minWithdrawal = decimal.Parse(environmentVariableMin, NumberStyles.AllowDecimalPoint, CultureInfo.InvariantCulture);
+        if (environmentVariableMax == null) _maxWithdrawal = 21_000_000;
```
cvicenti, line 448: Ideally these magic numbers should be constants at the top of the file so we can easily find them if we need to change them

src/Pages/Withdrawals.razor
markettes: Done by the IDE, MB I didnt payed attention to it

src/Pages/Withdrawals.razor
markettes: I will do it

PR #49, opened 2023-01-10
Fixed wrong succeed on job retries

RodriFS:


src/Helpers/JobTypes.cs
```diff
@@ -89,21 +89,19 @@ public static async Task Execute(IJobExecutionContext context, Func<Task> callba
 
         var trigger = context.Trigger as SimpleTriggerImpl;
 
-        if (trigger!.TimesTriggered >= intervals.Length)
+        if (trigger!.TimesTriggered <= intervals.Length)
         {
-            return;
+            var repeatInterval = intervals[trigger!.TimesTriggered - 1];
```
Jossec101, line 94: If the interval is a collection you can use collection.Last() or even Collection.LastOrDefault() instead of indexes

RodriFS: unfortunatelly, I'm not consuming the array, I'm just traversing it. Each time that it runs, it gets the trigger count and so it gets the next of the array.
Another way would be to consume the first of the array, then save the array in the job data map again. But last time I tried to update a value inside a job, it didn't work

Jossec101: LGTM

Jossec101 (approved):
LGTM

PR #50, opened 2023-01-11
Peer connection before opening a channel #47

### LINKED ISSUE #47: Peer connection is not done prior to opening a channel
In LND you need to add a peer prior to opening a channel, or use the —connect with the ‘openchannel’ command with the "pubkey@host:port" pattern but we are not doing it on the LightningService

Jossec101:
Fixed #47 by adding a peer connection prior to opening the channel, if the connection was already set we handle the potential exception

### DIFF
--- a/src/Services/LightningService.cs
+++ b/src/Services/LightningService.cs
@@ -249,11 +249,74 @@ namespace FundsManager.Services
                     NodePubkey = ByteString.CopyFrom(Convert.FromHexString(destination.PubKey)),
                 };
 
+                //Prior to opening the channel, we add the remote node as a peer
+                var remoteNodeInfo = await GetNodeInfo(channelOperationRequest.DestNode?.PubKey);
+                if (remoteNodeInfo == null)
+                {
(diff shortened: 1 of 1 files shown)


src/Services/LightningService.cs
```diff
@@ -249,11 +249,74 @@ public async Task OpenChannel(ChannelOperationRequest channelOperationRequest)
...
+                catch (RpcException e)
+                {
+                    if (!e.Message.Contains("already connected to peer"))
+                    {
+                        throw;
+                    }
+                    else
+                    {
+                        isPeerAlreadyConnected = true;
+                    }
```
RodriFS, line 294: what if the rpc error is that it couldn't establish the connection?

Jossec101: This is a try-catch inside of the big try-catch block, so any exception will pause the job and report the error, the thing about this specific try-catch is to handle a exception which should not be throw (peer already connected)

RodriFS (approved):

PR #58, opened 2023-01-12
Marking jobs as failed when attempts exeeded

RodriFS:


Jossec101 (commented):
We should do this also for WalletWithdrawalRequests

Jossec101 (approved):

PR #66, opened 2023-01-12
Fixed #65

### LINKED ISSUE #65: Withdrawals PSBTs do not have Global XPUBs field populated and the derivation paths for each input are missing
Withdrawals PSBTs do not have Global XPUBs field populated and the derivation paths for each input are missing. This comes from an initial thought that NodeGuard would never sign withdrawals, which is not valid anymore.

Jossec101:
Now withdrawals PSBTs will support the Global Xpubs field and each input will have its derivation path


markettes:
LGTM

markettes (approved):

PR #68, opened 2023-01-13
Fixed modal not cleared when closed

RodriFS:
We should be careful with the nesting of if statements and using the same modal for two different reasons, it becomes confusing to debug.

### DIFF
--- a/src/Helpers/Constants.cs
+++ b/src/Helpers/Constants.cs
@@ -1,0 +1,12 @@
+using System.Globalization;
+
+public class Constants
+{
+    public static readonly long MINIMUM_CHANNEL_CAPACITY_SATS;
+    
+    static Constants()
+    {
+        var environmentVariable = Environment.GetEnvironmentVariable("MINIMUM_CHANNEL_CAPACITY_SATS") ?? throw new InvalidOperationException("MINIMUM_CHANNEL_CAPACITY_SATS not set");
+        MINIMUM_CHANNEL_CAPACITY_SATS = long.Parse(environmentVariable, NumberStyles.AllowDecimalPoint, CultureInfo.InvariantCulture);
+    }
+}
--- a/src/Pages/ChannelRequests.razor
+++ b/src/Pages/ChannelRequests.razor
@@ -27,7 +27,8 @@
               ShowPageSizes="true"
               Editable="true"
               ShowValidationsSummary="true"
-              Striped="true">
+              Striped="true"
+              PopupClosing="OnChannelRequestRejectModalClosing">
         <PopupTitleTemplate>
             <h2>Create Channel Request</h2>
         </PopupTitleTemplate>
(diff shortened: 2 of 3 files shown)


src/Pages/ChannelRequests.razor
```diff
@@ -320,13 +321,13 @@
...
     // New Request integration
     private List<Wallet> _allWallets = new List<Wallet>();
-    protected DataGrid<ChannelOperationRequest> datagridRef;
+    protected DataGrid<ChannelOperationRequest>? datagridRef;
     private List<Node>? _manageableNodes;
     private int _selectedSourceNodeId;
     private Node? _selectedDestNode;
     private int _selectedWalletId;
     private string? _destNodeName;
-    private long _amount = 20000L;
```
markettes, line 329: Just out of curiosity, why did you remove this?

src/Pages/ChannelRequests.razor
```diff
@@ -574,9 +585,7 @@
 
     private void HideRejectOrCancelModal()
     {
-        _rejectCancelModalVisible = false;
-        _cancelOrRejectReason = null;
-
+        ResetChannelCancelModal();
```
markettes, line 588: I think you could just change the method called in the `Clicked` tag in line 286 and remove this function.

RodriFS: true, changing it. Thanks!

src/Pages/ChannelRequests.razor
RodriFS: Because to reset the modal I would have to again set the _amount variable to 20000L. 
If I duplicate this assignment, next time someone has to change it, it has to do it in 2 or more places.
Now I only assign it in the `ResetChannelRequestRejectModal` method, and then call that method on `OnInitializedAsync` and everywhere else that I needed.

Another option (maybe better) would be to create a "Constants" file, in which we store all this "magic numbers"

markettes: I think this might be the minimum channel capacity that is set as an env var. Maybe we could use that. `MINIMUM_CHANNEL_CAPACITY_SATS: 20000` is set in the docker compose file.

RodriFS: awesome, I'll use that

markettes (approved):
LGTM

src/Helpers/Constants.cs
```diff
@@ -0,0 +1,12 @@
+using System.Globalization;
+
+public class Constants
+{
+    public static readonly long MINIMUM_CHANNEL_CAPACITY_SATS;
+    
+    static Constants()
+    {
+        var environmentVariable = Environment.GetEnvironmentVariable("MINIMUM_CHANNEL_CAPACITY_SATS") ?? throw new InvalidOperationException("MINIMUM_CHANNEL_CAPACITY_SATS not set");
```
Jossec101, line 9: I like this, we can migrate step-by-step to this new approach

src/Helpers/Constants.cs
```diff
@@ -0,0 +1,12 @@
...
+        MINIMUM_CHANNEL_CAPACITY_SATS = long.Parse(environmentVariable, NumberStyles.AllowDecimalPoint, CultureInfo.InvariantCulture);
```
Jossec101, line 10: You should use tryparse, parse is dangerous and can throw exceptions

src/Helpers/Constants.cs
RodriFS: I'll create an issue

src/Helpers/Constants.cs
RodriFS: The idea is that it throws an exception, though. It would be nice to have this class throw before everything else is executed so you know that you forgot an env var right away

Jossec101 (approved):

src/Helpers/Constants.cs
Jossec101: Fair enough

PR #81, opened 2023-01-25
Fixed role based accesses

RodriFS:


src/Pages/ChannelRequests.razor
```diff
@@ -9,11 +9,18 @@
 
 <PageTitle>Channel Operation Requests</PageTitle>
 
-@if (!_hidePendingRequests)
```
Jossec101, line 12: What would be render if there are no requests, an empty datagrid?

src/Pages/Wallets.razor
```diff
@@ -284,8 +283,14 @@
...
     {
-
-        _wallets = await WalletRepository.GetAll();
+        if (ClaimsPrincipal != null && ClaimsPrincipal.IsInRole(ApplicationUserRole.FinanceManager.ToString()))
+        {
+            _wallets = await WalletRepository.GetAll();
+        }
+        else
+        {
+           _wallets = await WalletRepository.GetAvailableWallets();
```
Jossec101, line 292: What's the reason behind this if else?

src/Pages/Withdrawals.razor
```diff
@@ -355,59 +352,40 @@
...
-    private IWalletRepository WalletRepository { get; set; }
-
-    [Inject]
-    private IBitcoinService BitcoinService { get; set; }
+@inject IToastService ToastService
+@inject IWalletWithdrawalRequestRepository WalletWithdrawalRequestRepository
+@inject IWalletWithdrawalRequestPsbtRepository WalletWithdrawalRequestPsbtRepository
+@inject IWalletRepository WalletRepository
+@inject IBitcoinService BitcoinService
+@inject ISchedulerFactory SchedulerFactory
```
Jossec101, line 379: We can do injects in both ways, but it is true that we have not used the same approach always

src/Pages/ChannelRequests.razor
RodriFS: For finance managers will render the requests waiting for signature.
For node managers will render the requests the node manager created and are pending signature from the finance manager.
For admins it won't render at all. All requests will show in the other datagrid

The thing is that i needed the "new" button for node managers, but the button needs the datagrid to exist, so i had to separate them into pending and approved for node managers too

src/Pages/ChannelRequests.razor
```diff
@@ -9,11 +9,18 @@
...
+@if (_isFinanceManager || _isNodeManager)
 {
     <Field Flex="Flex.JustifyContent.Between">
+    @if (_isFinanceManager)
+    {
         <h3>Requests awaiting my signature</h3>
+    }
+    else if (_isNodeManager)
+    {
+        <h3>Requests awaiting signatyre by a Finance Manager</h3>
```
RodriFS, line 21: Oops, typo. will fix

src/Pages/Wallets.razor
RodriFS: For node managers and admins to not see pending wallets. I think is pointless.

src/Pages/Withdrawals.razor
RodriFS: There were a lot of null warnings on this, that's why I moved it.
I tried to remove all null warnings but some of them are literally impossible to remove. For example when you have a ref in a component

Jossec101: Yes null handling cannot be complete removed in C# even with modern features

Jossec101 (approved):
LGTM

PR #88, opened 2023-02-06
Unit testing services (Open Channel Method)

RodriFS:
First iteration for testing services in nodeguard

Screenshot 2023-02-06 at 17 03 36


src/Services/LightningService.cs
```diff
@@ -674,9 +682,7 @@ await foreach (var response in openStatusUpdateStream.ResponseStream.ReadAllAsyn
             {
```
RodriFS, line 682: Should this check that there are M of N signed instead of checking if there's one more than before?

src/Pages/ChannelRequests.razor
```diff
@@ -165,7 +165,7 @@
             </DataGridColumn>
             <DataGridCommandColumn TItem="ChannelOperationRequest" Caption="Actions" Displayable="true">
                 <EditCommandTemplate>
-                    <Button Color="Color.Primary" hidden=@_isNodeManager Clicked="@(() => ShowModal(context.Item))">Approve</Button>
+                    <Button Color="Color.Primary" hidden=@(!_isFinanceManager) Clicked="@(() => ShowModal(context.Item))">Approve</Button>
```
Jossec101, line 168: ? Why is this in this PR 🤔

src/Services/LightningService.cs
```diff
@@ -479,21 +401,6 @@ await foreach (var response in openStatusUpdateStream.ResponseStream.ReadAllAsyn
...
-                                        partialSigsCountAfterSignature <= partialSigsCount)
-                                    {
-                                        var invalidNoOfPartialSignatures =
-                                            $"Invalid expected number of partial signatures after signing the PSBT";
-
-                                        _logger.LogError(invalidNoOfPartialSignatures);
-                                        throw new ArgumentException(
-                                            invalidNoOfPartialSignatures);
-                                    }
-
```
Jossec101, line 496: Why is this block removed?

Jossec101 (commented):
A few things:

- Pentests unit tests should be removed, they are useless and fail
- We should modify .gitlab-ci.yml to include unit tests prior to building and not build if tests fail

Good work! Let's keep incrementing coverage step by step

src/Pages/ChannelRequests.razor
RodriFS: That's the small bug I caught that I mentioned. I can separate it to other PR if needed

src/Services/LightningService.cs
RodriFS: Because is repeated inside the SignPSBT function

src/Pages/ChannelRequests.razor
Jossec101: Its ok

.gitlab-ci.yml
```diff
@@ -39,10 +39,19 @@ build-release:
...
-    - "cd $test_path"
-    - "dotnet test"
-
+    - cd test/FundsManager.Tests
+    - dotnet test
+  artifacts:
+    paths:
+      - ./UnitTestResult.xml
+  rules:
+    - if: '$CI_COMMIT_REF_NAME == "develop" &&  $CI_PIPELINE_SOURCE != "merge_request_event"'
```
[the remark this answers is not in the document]
RodriFS: ok!

PR #92, opened 2023-02-13
Fix unit testing gitlab ci

RodriFS:
- Moved unit tests to before builds
- Added artifact for uploading results as a file


Jossec101 (approved):
LGTM

Issue #91, opened 2023-02-07
gRPC API


Jossec101:
NodeGuard needs a gRPC API for specific uses cases like:
- Getting information from Liquidity Rules associated to a node and its channels
- Generating wallet addresses for actions such as submarine swaps.
- Other: TBD

PR #95, opened 2023-02-14
Allow hot wallet creation

RodriFS:


src/Data/ApplicationDbContext.cs
```diff
@@ -47,6 +47,7 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
 
             modelBuilder.Entity<Node>().HasIndex(x => x.PubKey).IsUnique();
             modelBuilder.Entity<Node>().HasIndex(x => x.Name).IsUnique();
+            modelBuilder.Entity<Wallet>().HasIndex(x => x.InternalWalletSubDerivationPath).IsUnique();
```
Jossec101, line 50: I see a big issue here, the InternalWalletSubderivatonPath only holds 1 number, but it should be InternalWalletPath+SubPath, why? Because if there is a compromise and the internal wallet needs to change, the unique will make it impossible to start from 0 again.

src/Data/Repositories/Interfaces/IWalletRepository.cs
```diff
@@ -41,7 +41,7 @@ public interface IWalletRepository
 
     (bool, string?) RemoveRange(List<Wallet> types);
 
-    (bool, string?) Update(Wallet type, bool includeKeysUpdate = false);
+    Task<(bool, string?)> Update(Wallet type);
```
Jossec101, line 44: Why make the update a task? Updates are synchronous always.

src/Data/Repositories/WalletRepository.cs
```diff
@@ -134,18 +136,54 @@ public async Task<(bool, string?)> AddRangeAsync(List<Wallet> type)
...
-        public (bool, string?) Update(Wallet type, bool includeKeysUpdate = false)
-        {
+        public async Task<(bool, string?)> Update(Wallet type)
+        { 
             using var applicationDbContext = _dbContextFactory.CreateDbContext();
+            var wallet = await applicationDbContext.Wallets.Include(w => w.Keys).FirstOrDefaultAsync(x => x.Id == type.Id);
 
             type.SetUpdateDatetime();
-            if (!includeKeysUpdate)
-                type.Keys?.Clear();
```
Jossec101, line 143: Watchout, this is something that was added to avoid conflicts with EF Core, what's the purpose of this removal?

src/Data/Repositories/WalletRepository.cs
```diff
@@ -134,18 +136,54 @@ public async Task<(bool, string?)> AddRangeAsync(List<Wallet> type)
...
 
-            type.ChannelOperationRequestsAsSource?.Clear();
-            return _repository.Update(type, applicationDbContext);
+            bool wasHotWallet = wallet.IsHotWallet;
```
Jossec101, line 146:
```suggestion
            var wasHotWallet = wallet.IsHotWallet;
```

src/Data/Repositories/WalletRepository.cs
```diff
@@ -134,18 +136,54 @@ public async Task<(bool, string?)> AddRangeAsync(List<Wallet> type)
...
+            {
+                foreach (var key in type.Keys)
+                {
+                    if (!Key.Contains(wallet.Keys, key))
+                    {
+                        wallet.Keys.Add(key);
+                    }
+                }
+            }
+
```
Jossec101, line 167: This is related to the keys update? right?

src/Pages/Wallets.razor
```diff
@@ -42,45 +42,40 @@
                         <Button Color="Color.Success" TextColor="TextColor.Light" Clicked="@context.Clicked" Block>New</Button>
                     </NewCommandTemplate>
                     <EditCommandTemplate>
-                        <Button Color="Color.Primary" Clicked="@context.Clicked" Block>Edit</Button>
+                        <Button Color="Color.Primary" Clicked="@context.Clicked" @onclick:stopPropagation="true" Block>Edit</Button>
```
Jossec101, line 45: Using @events:action with Blazorise has given me lots of issues in the past 🤔

Jossec101 (commented):
Overall I would say to review how the subderivationpath is stored at db level, I would go for a full path in that case

src/Data/Repositories/Interfaces/IWalletRepository.cs
RodriFS: I'm doing a async call to the database on file `src/Data/Repositories/WalletRepository.cs` line 142

src/Data/Repositories/WalletRepository.cs
RodriFS: because we are now tracking the object instead of getting it from the argument, there is no conflict anymore. We tested it with all cases and it was not needed

src/Data/Repositories/WalletRepository.cs
RodriFS: exactly. If the user edits the wallet before finalising it and it makes it hot, we want to remove the user keys added.

src/Pages/Wallets.razor
RodriFS: this prevents the bug of the wallet keys displaying when you click a button. I can remove this if this causes problems, since is not a super annoying bug

src/Data/Repositories/Interfaces/IWalletRepository.cs
Jossec101: I mean, updates are sync in all the repos because EF Core Update operation is always sync, there's no need to overload with async if no async operation is really done

src/Pages/Wallets.razor
Jossec101: I would say maintain it if it works 👯

src/Services/BitcoinService.cs
```diff
@@ -68,7 +69,7 @@ public async Task<(decimal, long)> GetWalletConfirmedBalance(Wallet wallet)
 
             var (_, explorerClient) = LightningHelper.GenerateNetwork();
 
-            var balance = await explorerClient.GetBalanceAsync(wallet.GetDerivationStrategy());
+            var balance = await explorerClient.Execute(x => x.GetBalanceAsync(wallet.GetDerivationStrategy(), default));
```
cvicenti, line 72: what are these changes for?

RodriFS: The explorer client doesn't have an interface so we can't mock it. We added the [Unmockable](uhttps://github.com/riezebosch/Unmockablerl) library so we can do that.
One of the disadvantages is that you have to change production code

src/Jobs/ProcessNodeChannelAcceptorJob.cs
```diff
@@ -26,6 +26,7 @@
 using NBXplorer;
 using NBXplorer.DerivationStrategy;
 using Quartz;
+using Unmockable;
```
cvicenti, line 29: What does this do?

cvicenti: you answered it already

src/Services/BitcoinService.cs
cvicenti: nice thanks. I'm not too worried about that

cvicenti (approved):

src/Services/BitcoinService.cs
Jossec101: Anyway this will change in #91

PR #104, opened 2023-02-21
added Get Wallets method to Nodeguard services

RodriFS:


src/Proto/nodeguard.proto
```diff
@@ -58,4 +63,19 @@ message RequestWithdrawalResponse {
   //  optionalTransaction ID of the withdrawal
   string txid = 1;
   bool is_hot_wallet = 2;
+}
+
+message GetAvailableWalletsRequest {}
```
cvicenti, line 68: I wonder if we could generalise this method a little bit so we can add filters so we can use the same method instead of having to write future ones such as getAllHotWallets, getWalletbyId, getAllColdWallets, etc.
Food for thought

src/Rpc/NodeGuardService.cs
```diff
@@ -23,6 +23,9 @@ public interface INodeGuardService
         ServerCallContext context);
 
     Task<RequestWithdrawalResponse> RequestWithdrawal(RequestWithdrawalRequest request, ServerCallContext context);
+
+    Task<GetAvailableWalletsResponse>
+        GetAvailableWallets(GetAvailableWalletsRequest request, ServerCallContext context);
```
cvicenti, line 28: nit: I'd get rid of the extra newline here. Doesn't seem like the line is too long anyway

src/Proto/nodeguard.proto
```diff
@@ -58,4 +63,19 @@ message RequestWithdrawalResponse {
...
+
+message AccountKeySettings  {
+  string xpub = 3;
```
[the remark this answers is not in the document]
RodriFS: Forgot to change that. Thanks

src/Proto/nodeguard.proto
RodriFS: I needed both hot and cold wallets at the same time in the btcpayserver so I just did one call and filtered in the receiving end. But we can change it if you need it for your task

cvicenti (approved):

test/FundsManager.Tests/Rrpc/NodeGuardServiceTests.cs
```diff
@@ -0,0 +1,220 @@
...
+using Quartz;
+using Wallet = FundsManager.Data.Models.Wallet;
+
+namespace FundsManager.Rpc;
+
+public class NodeGuardServiceTests
+{
+    private readonly Random _random = new();
+
+    private (Mock<IDbContextFactory<ApplicationDbContext>>, ApplicationDbContext) SetupDbContextFactory()
```
Jossec101, line 17: Why mock the DbContextFactory? We can use mocks for the repositories so you don't need a in-memory db, this will make tests harder to maintain and probably slower

RodriFS: because I am also testing the repository that function leads to, which have a non intuitive Where clause
´´´
.Where(w => 
                    !w.IsArchived && !w.IsCompromised && w.IsFinalised &&
                    (type == WALLET_TYPE.Both || type == WALLET_TYPE.Cold && !w.IsHotWallet || type == WALLET_TYPE.Hot && w.IsHotWallet))
´´´

PR #105, opened 2023-02-21
Fixed unit tests

RodriFS:
Will stop compiling once we change the sub derivation path to int, but will be an easy fix


Jossec101 (approved):
LGTM

PR #107, opened 2023-02-22
Proposal of refactor of LightningService

RodriFS:
I made this so we can think about doing some more work on refactoring services.

The changes in question basically remove from the LightningService class all methods that are not part of the interface, and move them to a LightningServiceHelper class.
This allows us to maintain this helper functions public, since they should be private now but for testing purposes we needed them to be public.

This changes still violate the single responsibility principle on most methods, but it will help us start to move more stuff to the LightningServiceHelper so we can test smaller and smaller functions. From there we should separate this helper into classes related by responsibility.

All changes were on tested functions and tests still pass, so this should be straighforward

### DIFF
--- a/src/Pages/Shared/_Layout.cshtml
+++ b/src/Pages/Shared/_Layout.cshtml
@@ -26,7 +26,6 @@
 </head>
 <body>
     <div class="container">
-        <partial name="_CookieConsentPartial" optional />
         <main role="main" class="pb-1">
             @RenderBody()
         </main>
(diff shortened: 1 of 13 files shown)


src/Pages/Shared/_Layout.cshtml
```diff
@@ -26,7 +26,6 @@
 </head>
 <body>
     <div class="container">
-        <partial name="_CookieConsentPartial" optional />
```
RodriFS, line 29: also we can remove this because we're not using it and it's giving us an error in the IDE

markettes (approved):
LGTM

PR #109, opened 2023-02-22
Allow disabling autosweep on nodes

RodriFS:


src/Jobs/SweepNodeWalletsJob.cs
```diff
@@ -148,9 +148,17 @@ public async Task Execute(IJobExecutionContext context)
             _logger.LogError("{JobName} failed on node with id: {NodeId}, reason: node not found",
                 nameof(SweepNodeWalletsJob),
                 managedNodeId);
-            throw new ArgumentException("node not found", nameof(node));
+            await context.Scheduler.DeleteJob(context.JobDetail.Key, context.CancellationToken);
+            return;
```
RodriFS, line 152: jobs weren't cancelled when a node was deleted so it kept giving this error forever

src/Pages/Nodes.razor
```diff
@@ -179,7 +179,7 @@
 
     private async Task GetData()
     {
-        _nodes = await NodeRepository.GetAll();
+        _nodes = await NodeRepository.GetAllManagedByFundsManager();
```
RodriFS, line 182: Not managed nodes should not be visible in the Nodes view, but they should be selectable from the channel request view

src/Jobs/SweepNodeWalletsJob.cs
Jossec101: Nice catch

Jossec101 (approved):
LGTM

PR #113, opened 2023-02-25
102 liquidity rules UI

Jossec101:
First UI version for managing channel liquidity rules


src/Data/Repositories/LiquidityRuleRepository.cs
```diff
@@ -105,7 +111,7 @@ public async Task<ICollection<LiquidityRule>> GetByNodePubKey(string nodePubKey)
             .Include(x=> x.Wallet)
             .ThenInclude(x=> x.InternalWallet)
             .Include(x=> x.Channel)
-            .Where(x=> x.Node.PubKey == nodePubKey).ToList();
+            .Where(x=> x.Node.PubKey == nodePubKey && x.Channel.IsAutomatedLiquidityEnabled).ToList();
```
cvicenti, line 114: No need to change this now but I would have written a different method or perhaps change the name of this one (or even better, pass isAutomatedLiquidityEnabled as a boolean with the method signature) to reflect its new purpose. Someone might want to get nodes by pubkey regardless of it having automated liquidity enabled or not

src/FundsManager.csproj
```diff
@@ -19,10 +19,10 @@
   <ItemGroup>
     <PackageReference Include="AutoMapper.Extensions.Microsoft.DependencyInjection" Version="12.0.0" />
     <PackageReference Include="AwsSignatureVersion4" Version="4.0.0" />
-    <PackageReference Include="Blazored.Toast" Version="3.2.2" />
-    <PackageReference Include="Blazorise.Bootstrap" Version="1.1.5" />
-    <PackageReference Include="Blazorise.Components" Version="1.1.5" />
-    <PackageReference Include="Blazorise.Icons.FontAwesome" Version="1.1.5" />
+    <PackageReference Include="Blazored.Toast" Version="4.0.0" />
```
cvicenti, line 22: Ideally we should address this lib updates on separate PRs

src/Services/NBXplorerService.cs
```diff
@@ -104,7 +104,9 @@ public async Task<BroadcastResult> BroadcastAsync(Transaction tx, bool testMempo
   public async Task<StatusResult> GetStatusAsync(CancellationToken cancellation = default)
   {
     var client = await LightningHelper.CreateNBExplorerClient();
+
+    var statusResult = await client.GetStatusAsync(cancellation);
     
-    return await client.GetStatusAsync(cancellation);
+    return statusResult;
```
cvicenti, line 110: Why this extra variable?

cvicenti (approved):

src/Services/NBXplorerService.cs
Jossec101: Debugging is easier on UI 👯

src/FundsManager.csproj
Jossec101: Agree

src/Data/Repositories/LiquidityRuleRepository.cs
Jossec101: Agree but right now we would not need never a liquidity rule which is disabled in liquidator, so that's why is hardcoded

PR #115, opened 2023-02-27
Restore deriving for existing multisig wallets

RodriFS:


src/Data/Models/Wallet.cs
```diff
@@ -109,7 +109,7 @@ public class Wallet : Entity
             {
                 var bitcoinExtPubKeys = Keys.Select(x =>
                     {
-                        if (x.InternalWalletId != null)
+                        if (IsHotWallet && x.InternalWalletId != null)
```
markettes, line 112: Maybe here is only the first necessary
```suggestion
                        if (IsHotWallet)
```

src/Helpers/LightningHelper.cs
```diff
@@ -50,23 +50,24 @@ public static void RemoveDuplicateUTXOs(this UTXOChanges utxoChanges)
...
-            foreach (var key in keys)
+            foreach (var key in wallet.Keys)
             {
                 if (string.IsNullOrWhiteSpace(key.MasterFingerprint) || string.IsNullOrWhiteSpace(key.XPUB)) continue;
 
                 var bitcoinExtPubKey = new BitcoinExtPubKey(key.XPUB, nbXplorerNetwork);
                 var masterFingerprint = HDFingerprint.Parse(key.MasterFingerprint);
                 var rootedKeyPath = new RootedKeyPath(masterFingerprint, new KeyPath(key.Path));
-                if (key.InternalWalletId != null)
+                if (wallet.IsHotWallet && key.InternalWalletId != null)
```
markettes, line 70: Same in here
```suggestion
                if (wallet.IsHotWallet)
```

RodriFS:
Closing this because It's better if we redo it from the tested version

PR #119, opened 2023-03-01
Various fixes

Jossec101:
- Fixed mapperprofile
- Channels view async call fix
- BitcoinService now supports PSBTs for automated withdrawals of hot wallets
- Fixed NodeGuard service RPC call

### DIFF
--- a/src/Pages/Channels.razor
+++ b/src/Pages/Channels.razor
@@ -77,10 +77,17 @@
                 <DataGridColumn TItem="Channel" Field="@nameof(Channel.UpdateDatetime)" Caption="@nameof(Channel.UpdateDatetime).Humanize(LetterCasing.Sentence)" Sortable="false"/>
                 <DataGridColumn TItem="Channel" Caption="Channel Balance">
                     <DisplayTemplate>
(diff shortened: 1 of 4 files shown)


RodriFS (approved):

src/Pages/Channels.razor
```diff
@@ -77,10 +77,17 @@
...
-                        }</Text>
+                        @{
+                            var balance = Task.Run(()=> GetPercentageBalance(context)).Result;
+                            var percentageBalance = Convert.ToInt32(balance);
+                            var percentageBalanceRound = Math.Round(balance, 2);
+                        }
+                        <Progress Color="Color.Primary" Value="@percentageBalance"/>
+                        <Text>
+                            @{
+                                @($"{percentageBalanceRound}%")
```
markettes, line 88: That approach looks better than mine. But whats the difference of doing it like that and doing an await?

cvicenti (approved):

markettes (approved):

src/Pages/Channels.razor
Jossec101: @markettes Sometimes you cannot do async on a Razor code, so you need to use Task.Run() (never use .Result directly, it can lead to deadlocks or exceptions not handled).

PR #128, opened 2023-03-07
Added destination node to channels, added new methods to channels repo

Jossec101:
Added destination node to channels, added new methods to channels repository

### DIFF
--- a/src/Data/Repositories/ChannelRepository.cs
+++ b/src/Data/Repositories/ChannelRepository.cs
@@ -22,8 +22,12 @@ using FundsManager.Data.Models;
 using FundsManager.Data.Repositories.Interfaces;
 using FundsManager.Jobs;
 using FundsManager.Helpers;
-using Quartz;
+ using FundsManager.Services;
+ using Grpc.Core;
+ using Lnrpc;
+ using Quartz;
 using Microsoft.EntityFrameworkCore;
+ using Channel = FundsManager.Data.Models.Channel;
 
 namespace FundsManager.Data.Repositories
 {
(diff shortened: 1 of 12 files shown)


src/Data/Repositories/ChannelRepository.cs
```diff
@@ -22,8 +22,12 @@
 using FundsManager.Data.Repositories.Interfaces;
 using FundsManager.Jobs;
 using FundsManager.Helpers;
-using Quartz;
+ using FundsManager.Services;
+ using Grpc.Core;
+ using Lnrpc;
+ using Quartz;
 using Microsoft.EntityFrameworkCore;
+ using Channel = FundsManager.Data.Models.Channel;
```
RodriFS, line 30: I don't know why Rider adds this space before the "using", it happens to me as well

src/Data/Repositories/ChannelRepository.cs
```diff
@@ -175,5 +179,97 @@ public async Task<(bool, string?)> SafeRemove(Channel type, bool forceClose = fa
...
+        {
+            await using var applicationDbContext = await _dbContextFactory.CreateDbContextAsync();
+            
+            //check the channel exists in the node
+            var channelOperationRequest = channel.ChannelOperationRequests.FirstOrDefault();
+            
+            if (channelOperationRequest == null)
+            {
+                _logger.LogError("Error while marking as closed with id: {ChannelId}. No channel operation request found", channel.Id);
+                return (false, "No channel operation request found");
```
RodriFS, line 248: This is a very Golang thing to do, why don't we use custom exceptions?

RodriFS (approved):

src/Data/Repositories/ChannelRepository.cs
Jossec101: I took it from Golang long time ago, true 👯 . I mean exceptions are cool but they need to be try-catched and sometimes they are a pain in performance. The main problem/feature is that all of the repositories have a golang-based error approach.

RodriFS (approved):

src/Data/Repositories/ChannelRepository.cs
markettes: Happening to me too. It is a pain in the ***

PR #133, opened 2023-03-08
Channels UI changes:

Jossec101:
- Removed finance manager blank space for actions they should not do
- Fixed filtering and Open status is the default filter

### DIFF
--- a/src/Pages/Channels.razor
+++ b/src/Pages/Channels.razor
@@ -33,13 +33,14 @@
                             <DropdownToggle Color="Color.Primary">
                             </DropdownToggle>
                             <DropdownMenu>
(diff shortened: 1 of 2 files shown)


src/Pages/Channels.razor
```diff
@@ -623,6 +624,8 @@
 
     private bool OnStatusFilter(object itemValue, object searchValue)
     {
+    //If searchValue is null, we set it to the filter initial value in the field
+        searchValue ??= _statusFilter;
```
RodriFS, line 628: didn't know you could do that 🤯

RodriFS (approved):

PR #134, opened 2023-03-09
Add SQL-based migration to fix nodes info

Jossec101:
Add SQL-based migration to fix SourceNodeId and DestinationNodeId fields from Channels by looking at its requests


markettes (approved):
ChatGPT power xD
LGTM!

PR #140, opened 2023-03-14
117 rework closing of channels

markettes:
This branch adds the following features:
- Register the opening and closing of channels to one of our nodes made from outside of NodeGuard.
- Register the opening and closing of channels between two of our nodes made from outside of NodeGuard.


src/Pages/Nodes.razor
```diff
@@ -267,19 +272,55 @@
                 if (!result)
                 {
                     ToastService.ShowError($"{node.Name} could not be deleted");
+                    // IScheduler scheduler = await SchedulerFactory.GetScheduler();
+                    // scheduler.PauseJob(node.JobId.ToString());
+                    // scheduler.DeleteJob(node.JobId.ToString());
+
```
RodriFS, line 278: remove?

src/Data/Models/ChannelOperationRequest.cs
```diff
@@ -105,14 +105,14 @@ public decimal Amount
...
+        public int? WalletId { get; set; }
+        public Wallet? Wallet { get; set; }
 
         public int SourceNodeId { get; set; }
         public Node SourceNode { get; set; }
         public int? DestNodeId { get; set; }
         public Node? DestNode { get; set; }
-        public string UserId { get; set; }
+        public string? UserId { get; set; }
         public ApplicationUser User { get; set; }
```
Jossec101, line 116:
```suggestion
        public ApplicationUser? User { get; set; }
```

src/Jobs/NodeSubscriptorJob.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            var scheduler = await _schedulerFactory.GetScheduler();
+            
+            foreach (var managedNode in managedNodes)
+            {
+                if (managedNode.ChannelAdminMacaroon != null)
+                {
+                    var map = new JobDataMap();
+                    map.Put("nodeId", managedNode.Id.ToString());
+                    var job = RetriableJob.Create<NodeChannelSuscribeJob>(map, managedNode.Id.ToString(), new []{1,1,1,1});
```
RodriFS, line 36: should this be a retriable job? Shouldn't this run forever? I think you should just restart the job like in SweepAllNodesWalletsJob, removing the "false" flag from `JobExecutionException`

src/Jobs/NodeChannelSuscribeJob.cs
```diff
@@ -0,0 +1,187 @@
...
+                    _logger.LogInformation("The node: {@Node} is no longer ready to be supported quartz jobs", node);
+                    return;
+                }
+                
+                try {
+                    var channelEventUpdate = result.ResponseStream.Current;
+
+                    if (channelEventUpdate.OpenChannel != null)
+                    {
+                        if (String.IsNullOrEmpty(channelEventUpdate.OpenChannel.CloseAddress))
```
Jossec101, line 90: Is this really possible to happen?

src/Jobs/NodeChannelSuscribeJob.cs
```diff
@@ -0,0 +1,187 @@
...
+                        var channelOpened = channelEventUpdate.OpenChannel;
+                        var fundingTxAndIndex = channelOpened.ChannelPoint.Split(":");
+                        var channel = new Channel()
+                        {
+                            ChanId = channelOpened.ChanId,
+                            SatsAmount = channelOpened.Capacity,
+                            Status = Channel.ChannelStatus.Open,
+                            IsAutomatedLiquidityEnabled = false,
+                            BtcCloseAddress = channelOpened.CloseAddress,
+                            FundingTx = fundingTxAndIndex[0],
```
Jossec101, line 103: You can use fundingTxAndIndex.First() probably

src/Pages/Nodes.razor
```diff
@@ -267,19 +272,55 @@
...
+            await scheduler.ScheduleJob(job.Job, job.Trigger);
+
+            node.JobId = job.Job.Key.ToString();
+
+            var jobUpdateResult = NodeRepository.Update(node);
+
+            map = new JobDataMap();
+            map.Put("managedNodeId", node.Id.ToString());
+            
+            var acceptorJob = RetriableJob.Create<ProcessNodeChannelAcceptorJob>(map, node.Id.ToString(), retryList);
```
RodriFS, line 314: do we need this? I think is already setup in Program.cs

src/Jobs/NodeChannelSuscribeJob.cs
```diff
@@ -0,0 +1,187 @@
...
+                            _logger.LogInformation("Channel with id: {ChannelId} added to the system", channel.Id);
+                        }
+                        else
+                        {
+                            _logger.LogInformation("Channel with id: {ChannelId} already exists in the system", channel.Id);
+                        }
+                        
+                    }
+
+                    if (channelEventUpdate.ClosedChannel != null)
```
Jossec101, line 153: Perhaps the channelEventUpdate types should be a switch instead of different ifs in the code

src/Jobs/NodeChannelSuscribeJob.cs
```diff
@@ -0,0 +1,187 @@
...
+    public NodeChannelSuscribeJob(ILogger<NodeChannelSuscribeJob> logger, ILightningService lightningService, INodeRepository nodeRepository, IChannelRepository channelRepository, ISchedulerFactory schedulerFactory)
+    {
+        _logger = logger;
+        _lightningService = lightningService;
+        _nodeRepository = nodeRepository;
+        _channelRepository = channelRepository;
+        _schedulerFactory = schedulerFactory;
+    }
+        
+    public async Task Execute(IJobExecutionContext context)
```
Jossec101, line 54: This Jobs needs some unit tests.

src/Pages/Nodes.razor
```diff
@@ -267,19 +272,55 @@
...
+    {
+        try
+        {
+            IScheduler scheduler = await SchedulerFactory.GetScheduler();
+
+            var map = new JobDataMap();
+            map.Put("nodeId", node.Id);
+
+            var retryList = RetriableJob.ParseRetryListFromString(Constants.JOB_RETRY_INTERVAL_LIST_IN_MINUTES);
+            var job = RetriableJob.Create<NodeChannelSuscribeJob>(map, node.Id.ToString(), retryList);
```
RodriFS, line 304: Same here, if it's a "forever" job, you don't need to set it up every time

RodriFS (dismissed):

src/Jobs/NodeSubscriptorJob.cs
```diff
@@ -0,0 +1,52 @@
...
+                    await scheduler.ScheduleJob(job.Job, job.Trigger);
```
Jossec101, line 37: Will this job refire if any exception happens?

src/Pages/Nodes.razor
```diff
@@ -198,6 +202,7 @@
             {
                 ToastService.ShowSuccess($"Node {arg.Item.Name} Created");
                 _nodes = await NodeRepository.GetAll();
+                CreateJobs(arg.Item);
```
Jossec101, line 205:
```suggestion
                await CreateJobs(arg.Item);
```

Jossec101 (commented):
- The job logic should be at least unit tested
- Added a few comments

src/Jobs/NodeSubscriptorJob.cs
markettes: Done

src/Pages/Nodes.razor
markettes: Done

src/Data/Models/ChannelOperationRequest.cs
markettes: Done

src/Jobs/NodeChannelSuscribeJob.cs
markettes: If the channelacceptor for some reason fails that could happen. Actually it was happening to me before I fixed the ChannelAcceptor init. It isnt usually null but an empty string.

markettes: It also happens when you open from outside of NG a channel from two registered nodes. It detects two events, one with that field empty.

src/Jobs/NodeChannelSuscribeJob.cs
markettes: Should be the same right? Is that way more secure or better?

src/Jobs/NodeChannelSuscribeJob.cs
markettes: Yeah, probably. It was only two types of event that we wanted so thats why I did an if

src/Jobs/NodeChannelSuscribeJob.cs
markettes: On my way!

src/Pages/Nodes.razor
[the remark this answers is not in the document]
markettes: Done :D

src/Jobs/NodeChannelSuscribeJob.cs
```diff
@@ -0,0 +1,187 @@
...
+            var node = await _nodeRepository.GetById(nodeId);
+            
+            if (node == null)
+            {
+                _logger.LogInformation("The node: {@Node} is no longer ready to be supported quartz jobs", node);
+                return;
+            }
+
+            var client = LightningService.CreateLightningClient(node.Endpoint);
+            var result = client.Execute(x => x.SubscribeChannelEvents(new ChannelEventSubscription(),
```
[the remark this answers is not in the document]
markettes: Nope

src/Jobs/NodeChannelSuscribeJob.cs
markettes: Done

src/Jobs/NodeChannelSuscribeJob.cs
markettes: Done

src/Jobs/NodeSubscriptorJob.cs
Jossec101: @RodriFS If i am not wrong, this wont refire right?

src/Services/LightningService.cs
```diff
@@ -1222,8 +1222,11 @@ public async Task<(long?, long?)> GetChannelBalance(Channel channel)
             var chan = result.Channels.FirstOrDefault(x => x.ChanId == channel.ChanId);
             if(chan == null)
                 return (null, null);
+            
+            var htlcsLocal = chan.PendingHtlcs.Where(x => x.Incoming == true).Sum(x => x.Amount);
+            var htlcsRemote = chan.PendingHtlcs.Where(x => x.Incoming == false).Sum(x => x.Amount);
```
Jossec101, line 1227: @markettes There's a chan.UnsettledBalance field with that info

Jossec101 (changes requested):

src/Jobs/NodeSubscriptorJob.cs
markettes: Now it is a simple Job and if you go into the job whenever it throws an exception it is refired automatically

src/Services/LightningService.cs
markettes: Thats true, but you cant get to know in what direction is that unsettled balance flowing

PR #146, opened 2023-04-03
Adds support for Settlements on BTCPayServer:

cvicenti:
* New field RequestMetadata to be set when the request contains additional information
* for the time being it will not be shown in NG views


markettes (approved):
LGTM

RodriFS (approved):

PR #157, opened 2023-04-25
RPC methods for opening, closing channels, adding and getting nodes.

Jossec101:
feat: add mapping from Node to Nodeguard.Node
feat: add GetAllManagedByNodeGuard method to INodeRepository
refactor: replace GetAllManagedByFundsManager with GetAllManagedByNodeGuard in jobs
refactor: replace ReturningFundsMultisigWallet with ReturningFundsWallet in Node model and repositories
refactor: replace ReturningFundsMultisigWalletId with ReturningFundsWalletId in Node model
refactor: update ProcessNodeChannelAcceptorJob to use ReturningFundsWallet instead of ReturningFundsMultisigWallet
refactor: update SweepAllNodesWalletsJob to use GetAllManagedByNodeGuard instead of GetAllManagedByFundsManager
refactor: update SweepNodeWalletsJob to use ReturningFundsWallet instead of ReturningFundsMultisigWallet
    refactor(Pages): rename NodeRepository.GetAllManagedByFundsManager to NodeRepository.GetAllManagedByNodeGuard
feat: Added AddNodes, Getodes, OpenChannel and CloseChannel to gRPC API
test: Added testing for the above new RPC methods
feat: New migration to support renaming of Nodes returning funds property

### DIFF
--- a/src/Rpc/NodeGuardService.cs
+++ b/src/Rpc/NodeGuardService.cs
@@ -11,6 +11,7 @@ using NBXplorer.DerivationStrategy;
 using Nodeguard;
 using Quartz;
 using LiquidityRule = Nodeguard.LiquidityRule;
+using Node = Nodeguard.Node;
 using Wallet = FundsManager.Data.Models.Wallet;
 
 namespace FundsManager.Rpc;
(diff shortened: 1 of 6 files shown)


src/Rpc/NodeGuardService.cs
```diff
@@ -250,4 +271,201 @@ ISchedulerFactory schedulerFactory
...
+        }
+        catch (Exception e)
+        {
+            _logger?.LogError(e, "Error getting nodes through gRPC");
+
+            throw new RpcException(new Status(StatusCode.Internal, e.Message));
+        }
+    }
+
+    public override async Task<OpenChannelResponse> OpenChannel(OpenChannelRequest request, ServerCallContext context)
```
RodriFS, line 337: I think we might need to refactor in the future, This code is repeated in some other parts of the app

Jossec101: Invoking the job? Yeah, it can be refactored

PR #160, opened 2023-04-30
BIP39 wallet import

Jossec101:
Added support for BIP39 hot wallets.

### DIFF
--- a/src/Data/Models/Wallet.cs
+++ b/src/Data/Models/Wallet.cs
@@ -63,6 +63,12 @@ namespace FundsManager.Data.Models
         /// Used to mark this wallet as imported from a BIP39 mnemonic though the seedphrase is not stored in the database
         /// </summary>
         public bool IsBIP39Imported { get; set; }
+
+        /// <summary>
+        /// Single sig wallet is a wallet that does not require any other signer to sign the transaction
+        /// </summary>
+        [NotMapped] 
+        public bool IsSingleSig => MofN == 1 && (IsHotWallet || IsBIP39Imported);
         
         /// <summary>
         /// Only when the remote signer is disabled, the seedphrase is stored in the database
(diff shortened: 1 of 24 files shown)


src/Data/Models/Wallet.cs
```diff
@@ -86,9 +102,9 @@ public class Wallet : Entity
         
 
         /// <summary>
-        /// The internal wallet is used to co-sign with other keys of the wallet entity
+        /// The internal wallet is used to co-sign with other keys of the wallet entity. It is optional for imported BIP39 wallets
         /// </summary>
-        public int InternalWalletId { get; set; }
+        public int? InternalWalletId { get; set; }
 
         public InternalWallet InternalWallet { get; set; }
```
markettes, line 109: Shouldnt this be optional aswell?

src/Services/LightningService.cs
```diff
@@ -712,7 +712,7 @@ public static (Node, Node) CheckNodesAreValid(ChannelOperationRequest channelOpe
                     errorParsingSubderivationPath);
             }
 
-            //We need to SIGHASH_ALL all inputs/outputs as fundsmanager to protect the tx from tampering by adding a signature
+            //We need to SIGHASH_ALL all inputs/outputs as  to protect the tx from tampering by adding a signature
```
markettes, line 715: Something is missing

test/FundsManager.Tests/Services/LightningServiceTests.cs
```diff
@@ -800,13 +1039,13 @@ public async Task OpenChannel_SuccessMultiSig()
...
-            
+
             //TODO Remove hack
             LightningService.CreateLightningClient = originalCreateLightningClient;
         }
-        
+
         [Fact]
-        public async Task OpenChannel_SuccessSingleSig()
+        public async Task OpenChannel_SuccesSingleSig()
```
markettes, line 1048:
```suggestion
        public async Task OpenChannel_SuccessSingleSig()
```

src/Data/Repositories/WalletRepository.cs
```diff
@@ -253,5 +269,121 @@ public async Task<(bool, string?)> FinaliseWallet(Wallet selectedWalletToFinalis
...
+                {
+                    return (false, "Error while getting the derivation scheme");
+                }
+
+                //Track wallet
+                await _nbXplorerService.TrackAsync(derivationStrategyBase, default);
+
+                //Since already existing wallet's utxos are not tracked by NBXplorer, we need to rescan the UTXO set for this wallet
+                //This is a long running operation in nbxplorer and should be queried in the background
+                await _nbXplorerService.ScanUTXOSetAsync(derivationStrategyBase, 1000, 30000, null, default);
```
markettes, line 377: If this is a long running operation, should we create a job for this? Because the await is going to make it stop for a while

src/Data/Models/Wallet.cs
Jossec101: ? Was added to inform the DB model for the FK to be nullable, but true

Jossec101:
```suggestion
        public InternalWallet? InternalWallet { get; set; }
```

src/Services/LightningService.cs
Jossec101:
```suggestion
            //We need to SIGHASH_ALL all inputs/outputs to protect the tx from tampering by adding a signature
```

src/Data/Repositories/WalletRepository.cs
Jossec101: The long running operation is on nbxplorer, not our side @markettes

markettes: Perfect, thanks!

markettes (approved):

PR #162, opened 2023-05-01
Added private channel check + model changes

Jossec101:
feat: Added private channel check for creating private channels outside the API
feat: add IsPrivate property to Channel model and set it in NodeChannelSubscribeJob and LightningService


markettes (approved):
LGTM

PR #165, opened 2023-05-03
bug mainnet start with embedded signer

Jossec101:
When starting a nodeguard instance from an empty db with the embedded signer crashes due to a bad null handling.


RodriFS (approved):

markettes (approved):
LGTM

PR #177, opened 2023-05-12
Channels being duplicated when opened from outside NG

markettes:


src/Data/Repositories/ChannelRepository.cs
```diff
@@ -62,12 +62,12 @@ IMapper mapper
...
             return await applicationDbContext.Channels.Include(x => x.ChannelOperationRequests).FirstOrDefaultAsync(x => x.Id == id);
         }
-        
+
         public async Task<Channel?> GetByChanId(ulong chanId)
         {
             await using var applicationDbContext = await _dbContextFactory.CreateDbContextAsync();
 
-            return await applicationDbContext.Channels.Include(x => x.ChannelOperationRequests).FirstOrDefaultAsync(x => x.ChanId == chanId);
+            return await applicationDbContext.Channels.Include(x => x.ChannelOperationRequests).FirstOrDefaultAsync(x => x.ChanId.Equals(chanId));
```
Jossec101, line 70: The == operator is ok in this case, what's the goal of using equals?

Jossec101 (approved):

src/Data/Repositories/ChannelRepository.cs
markettes: I was just discarding possible errors

PR #190, opened 2023-06-05
Added fee rate checks

Jossec101:
feat(Constants.cs): add MIN_SAT_PER_VB_RATIO, MAX_SAT_PER_VB_RATIO, and MAX_TX_FEE_RATIO for fee rate validation
feat(LightningService.cs): implement fee rate validation for finalized PSBT using new constants to ensure fees are within acceptable range

### DIFF
--- a/src/Services/LightningService.cs
+++ b/src/Services/LightningService.cs
@@ -194,9 +194,9 @@ namespace FundsManager.Services
             {
                 // 8 value + 1 script pub key size + 34 script pub key hash (Segwit output 2-0f-2 multisig)
                 var outputVirtualSize = combinedPSBT.GetGlobalTransaction().GetVirtualSize() + 43;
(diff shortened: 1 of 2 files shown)


src/Services/LightningService.cs
```diff
@@ -473,7 +473,38 @@ await foreach (var response in openStatusUpdateStream.ResponseStream.ReadAllAsyn
                                     finalizedPSBT.AssertSanity();
 
                                     channelfundingTx = finalizedPSBT.ExtractTransaction();
-
+                                    
+                                    //We check the feerate of the finalized PSBT by checkin a minimum and maximum allowed and also a fee-level max check in ratio
```
Jossec101, line 477:
```suggestion
                                    //We check the feerate of the finalized PSBT by checking a minimum and maximum allowed and also a fee-level max check in ratio
```

markettes (approved):
LGTM

PR #197, opened 2023-06-19
Add copy button to address modal and implement copy function

markettes:
- Add copy address button to address modal
- Fix async method call in Keys.cs
- Format auto corrections

### DIFF
--- a/src/Pages/Keys.razor
+++ b/src/Pages/Keys.razor
@@ -22,7 +22,7 @@
                   RowRemoving="OnRowRemoving"
                   RowUpdated="OnRowUpdated"
                   NewItemDefaultSetter="NewItemDefaultSetter">
-            
+
             <PopupTitleTemplate>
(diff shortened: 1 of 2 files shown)


src/Pages/Keys.razor
```diff
@@ -58,7 +58,7 @@
                         {
                             @StringHelper.TruncateHeadAndTail(context.XPUB, 6);
                             <span>&nbsp</span>
-                            <Button Color="Color.Primary" Clicked="@(() => CopyStrToClipboard(context.XPUB))" Size="Size.Small" Outline>
+                            <Button Color="Color.Primary" Clicked="@(async () => await CopyStrToClipboard(context.XPUB))" Size="Size.Small" Outline>
```
RodriFS, line 61: Do you really need to change this?

The => implies a return in the line. So you're actually returning a Task.

RodriFS (dismissed):

RodriFS (approved):

PR #218, opened 2023-06-28
Hide columns dropdown

RodriFS:
- Changed order of Source and Remote node columns in the first table to match the table below
- Dropdowns are implemented in the "Filters" row most of the time, except in the channel requests table because it has the new button in the top right instead of the top left and was too much of a change to do here.
- Dropdown doesn't hide on focus lost, same thing that happens with the hot wallet filters


markettes (approved):
LGTM super good idea

PR #225, opened 2023-07-04
Show user channel opening failure reason

RodriFS:
- Added logs in UI for "peer not online" and remote canceling funding
- Added Not retrying in case that the remote node cancels funding
- Added a directory inside "Pages" called "ChannelRequestsComponents" where we would put all non-shared components that are used in the ChannelRequests page.

### DIFF
--- a/src/Jobs/ChannelOpenJob.cs
+++ b/src/Jobs/ChannelOpenJob.cs
@@ -60,6 +60,8 @@ public class ChannelOpenJob : IJob
         }
         catch (Exception e)
         {
+            var shouldRetry = await LogToChannelRequest(openRequestId, e, context);
+
             await RetriableJob.OnFail(context, async () =>
             {
                 var request = await _channelOperationRequestRepository.GetById(openRequestId);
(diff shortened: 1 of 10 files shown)


src/Jobs/ChannelOpenJob.cs
```diff
@@ -68,9 +70,49 @@ public async Task Execute(IJobExecutionContext context)
             });
 
             _logger.LogError(e, "Error on {JobName}", nameof(ChannelOpenJob));
+            if (!shouldRetry)
+            {
+                await context.Scheduler.DeleteJob(context.JobDetail.Key, context.CancellationToken);
+            }
```
Jossec101, line 76: What if we implement a retry feature? wouldn't this kill that possibility?

Jossec101 (approved):

src/Jobs/ChannelOpenJob.cs
RodriFS: This only kills the retry for the specific case that the channel rejects funding. Because we know there's no point in retrying. Every other error still goes through the retry intervals 1,2,5,10,20

PR #235, opened 2023-07-14
Check upfront shutdown script opt

RodriFS:
I was unable to test this locally since doing `lncli getnodeinfo` on an eclair wallet doesn't show the features. Tested with c-lightning and it does show features, but c-lightning supports closing script.
Screenshot 2023-07-14 at 17 56 54

### DIFF
--- a/src/Services/LightningService.cs
+++ b/src/Services/LightningService.cs
@@ -255,8 +255,8 @@ namespace FundsManager.Services
                 if (upfrontShutdownScriptOpt && remoteNodeInfo.Features[(uint)FeatureBit.UpfrontShutdownScriptOpt] is { IsKnown: true } ||
                     upfrontShutdownScriptReq && remoteNodeInfo.Features[(uint)FeatureBit.UpfrontShutdownScriptReq] is { IsKnown: true })
                 {
-                    var keyPathInformation = await GetCloseAddress(channelOperationRequest, derivationStrategyBase, _nbXplorerService, _logger);
-                    closeAddress = keyPathInformation?.Address?.ToString();
+                    var address = await GetCloseAddress(channelOperationRequest, derivationStrategyBase, _nbXplorerService, _logger);
+                    closeAddress = address.Address.ToString();
                     openChannelRequest.CloseAddress = closeAddress;
                 }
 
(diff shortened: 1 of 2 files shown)


Jossec101:
Im missing a unit test here :D @RodriFS

src/Services/LightningService.cs
```diff
@@ -250,8 +248,20 @@ public async Task OpenChannel(ChannelOperationRequest channelOperationRequest)
...
 
+                // Check features to see if we need or is allowed to add a close address
+                var upfrontShutdownScriptOpt = remoteNodeInfo.Features.ContainsKey((uint)FeatureBit.UpfrontShutdownScriptOpt);
+                var upfrontShutdownScriptReq = remoteNodeInfo.Features.ContainsKey((uint)FeatureBit.UpfrontShutdownScriptReq);
+                string? closeAddress = null;
+                if (upfrontShutdownScriptOpt && remoteNodeInfo.Features[(uint)FeatureBit.UpfrontShutdownScriptOpt] is { IsKnown: true } ||
+                    upfrontShutdownScriptReq && remoteNodeInfo.Features[(uint)FeatureBit.UpfrontShutdownScriptReq] is { IsKnown: true })
+                {
+                    var keyPathInformation = await GetCloseAddress(channelOperationRequest, derivationStrategyBase, _nbXplorerService, _logger);
+                    closeAddress = keyPathInformation?.Address?.ToString();
```
Jossec101, line 259: Maybe here check that it is not null the closeAddress?

Jossec101 (changes requested):
I would say this class is important enough to any change we introduce be well tested via unit tests

src/Services/LightningService.cs
```diff
@@ -368,7 +378,7 @@ await foreach (var response in openStatusUpdateStream.ResponseStream.ReadAllAsyn
                                     CreationDatetime = DateTimeOffset.Now,
                                     FundingTx = fundingTx,
                                     FundingTxOutputIndex = response.ChanOpen.ChannelPoint.OutputIndex,
-                                    BtcCloseAddress = closeAddress?.Address.ToString(),
```
Jossec101, line 371: There's a possibility the closeAddress is null, we should handle it like the comment above.

src/Services/LightningService.cs
RodriFS: `GetCloseAddress` throws if the close address is null

RodriFS: `GetCloseAddress` throws if there is no close address. The signature of the function is wrong, it should return Task<KeyPathInformation> instead of Task<KeyPathInformation?>

RodriFS:
> Im missing a unit test here :D @RodriFS

Will do!

Jossec101 (approved):

PR #237, opened 2023-07-18
feat: filtering channels by funding wallet

markettes:


src/Data/Models/Channel.cs
```diff
@@ -44,31 +46,43 @@ public enum ChannelStatus
...
+        {
+            get
+            {
+                if (ChannelOperationRequests != null && ChannelOperationRequests.Count > 0)
+                    return ChannelOperationRequests.FirstOrDefault().Wallet.Id;
+                else
+                    return null;
+            }
+        }
+
```
RodriFS, line 76: would this work? 🤔
```
ChannelOperationRequests?.FirstOrDefault()?.Wallet?.Id;
```

src/Pages/Channels.razor
```diff
@@ -692,6 +715,19 @@
...
+        {
+            return true;
+        }
+        if (itemValue == null) return false;
+        if ((int)itemValue == (int)searchValue)
+        {
+            return true;
+        }
+        return false;
+    }
```
RodriFS, line 730: This should be the same, right?
```
return searchValue == null || (int)searchValue == 0 || (int)itemValue == (int)searchValue
```

RodriFS (approved):

src/Data/Models/Channel.cs
markettes: Its true, good point!

src/Pages/Channels.razor
```diff
@@ -93,7 +92,27 @@
...
+                        </Select>
+                    </FilterTemplate>
+                    <DisplayTemplate>
+                        @{
+                            Wallet wallet = null;
+                            if (context.OpenedWithId != null)
+                            {
+                                wallet = Task.Run(() => WalletRepository.GetById((int)context.OpenedWithId)).Result;
+                            }
+                            @(wallet == null ? "Unknown" : wallet.Name)
```
Jossec101, line 112: I wonder how much will this make the channels view slower, its async but maybe slows down, we already have the problem with the balance, worth researching

src/Pages/Channels.razor
```diff
@@ -692,6 +715,11 @@
         return true;
     }
 
+    private bool OnWalletFilter(object itemValue, object searchValue)
+    {
+        if (itemValue == null) return false;
```
RodriFS, line 720: Do you really need this line?

RodriFS (approved):

src/Pages/Channels.razor
markettes: yes, because on some cases you get null on the itemvalue

PR #240, opened 2023-07-20
Simultaneous operations in channels with same source and destination

RodriFS:
- We allow simultaneous channel close operations for same source and destination nodes
- Added feature flag for when we modify LND we can allow also the opening of channels with same source and destination nodes

### DIFF
--- a/src/Data/Repositories/ChannelRepository.cs
+++ b/src/Data/Repositories/ChannelRepository.cs
@@ -145,7 +145,7 @@ namespace FundsManager.Data.Repositories
 
             if (!closeRequestAddResult.Item1)
             {
-                _logger.LogError("Error while saving close request for channel with id: {RequestId}: " + closeRequestAddResult.Item2, type.Id);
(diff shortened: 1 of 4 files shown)


src/Data/Repositories/ChannelRepository.cs
```diff
@@ -145,7 +145,7 @@ public async Task<(bool, string?)> SafeRemove(Channel type, bool forceClose = fa
 
             if (!closeRequestAddResult.Item1)
             {
-                _logger.LogError("Error while saving close request for channel with id: {RequestId}", type.Id);
+                _logger.LogError("Error while saving close request for channel with id: {RequestId}: " + closeRequestAddResult.Item2, type.Id);
```
markettes, line 148: Could it be something like this?
```suggestion
                _logger.LogError("Error while saving close request for channel with id: {RequestId}: {error}", closeRequestAddResult.Item2, type.Id);
```

PR #242, opened 2023-07-24
Monitor channels job

RodriFS:
The idea for this PR is to be able to check for and add ghost channels to the NodeGuard

### DIFF
--- a/src/Helpers/Constants.cs
+++ b/src/Helpers/Constants.cs
@@ -59,7 +59,7 @@ public class Constants
 
     // Crons & Jobs
     public static readonly string MONITOR_WITHDRAWALS_CRON = "10 0/5 * * * ?";
-    public static readonly string MONITOR_CHANNELS_CRON = "0 0 0 * * ?";
+    public static readonly string MONITOR_CHANNELS_CRON = "0 0 */1 * * ?";
     public static readonly string JOB_RETRY_INTERVAL_LIST_IN_MINUTES = "1,2,5,10,20";
 
 
--- a/src/Helpers/Constants.cs
+++ b/src/Helpers/Constants.cs
@@ -55,6 +55,7 @@ public class Constants
 
     // Crons & Jobs
     public static readonly string MONITOR_WITHDRAWALS_CRON = "10 0/5 * * * ?";
+    public static readonly string MONITOR_CHANNELS_CRON = "0 0 0 * * ?";
     public static readonly string JOB_RETRY_INTERVAL_LIST_IN_MINUTES = "1,2,5,10,20";
 
 
@@ -159,6 +160,8 @@ public class Constants
 
         // Crons & Jobs
         MONITOR_WITHDRAWALS_CRON = Environment.GetEnvironmentVariable("MONITOR_WITHDRAWALS_CRON") ?? MONITOR_WITHDRAWALS_CRON;
+
+        MONITOR_CHANNELS_CRON = Environment.GetEnvironmentVariable("MONITOR_CHANNELS_CRON") ?? MONITOR_CHANNELS_CRON;
 
         JOB_RETRY_INTERVAL_LIST_IN_MINUTES = Environment.GetEnvironmentVariable("JOB_RETRY_INTERVAL_LIST_IN_MINUTES") ?? JOB_RETRY_INTERVAL_LIST_IN_MINUTES;
 
--- a/src/Jobs/ChannelMonitorJob.cs
+++ b/src/Jobs/ChannelMonitorJob.cs
@@ -61,7 +61,7 @@ public class ChannelMonitorJob : IJob
 
             if (node1 == null)
             {
-                _logger.LogInformation("The node: {@Node} is no longer ready to be supported quartz jobs", node1);
+                _logger.LogInformation("The node {NodeId} was set up for monitoring but the node doesn't exist anymore", nodeId);
                 return;
             }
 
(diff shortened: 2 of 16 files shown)


src/Jobs/ChannelMonitorJob.cs
```diff
@@ -0,0 +1,163 @@
...
+                    {
+                        Name = foundNode.Alias,
+                        PubKey = foundNode.PubKey,
+                    };
+                    var addNode = await _nodeRepository.AddAsync(node2);
+                    if (!addNode.Item1)
+                    {
+                        throw new Exception(addNode.Item2);
+                    }
+                }
```
daliclovr, line 96: I think this logic of creating a new node if not found in the repository should be extracted, maybe creating a method in the repository if makes sense that actually creates a new one passing a lightning client? Or at least extracting it outside of this function if it doesn't makes sense in that scope, what do you think?

src/Jobs/ChannelMonitorJob.cs
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        _logger.LogInformation("{JobName} ended", nameof(ChannelMonitorJob));
+    }
+
+    private async Task RecoverGhostChannels(Node source, Node destination, Channel? channel)
+    {
+        if (!channel.Initiator) return;
+
+        await using var context = await _dbContextFactory.CreateDbContextAsync();
```
daliclovr, line 124: i would chose a better name such as `dbContext` 😄

src/Services/LightningService.cs
```diff
@@ -635,6 +562,96 @@ await foreach (var response in openStatusUpdateStream.ResponseStream.ReadAllAsyn
...
+            if (channelUpdate.Item1 == false)
+            {
+                _logger.LogError(
+                    "Could not assign channel id to channel operation request: {RequestId} reason: {Reason}",
+                    channelOperationRequest.Id,
+                    channelUpdate.Item2);
+            }
+        }
+
+        public static async Task<Channel> CreateChannel(Node source, int destId, ChannelPoint channelPoint, long satsAmount, string? closeAddress = null)
```
daliclovr, line 620: Is this the best place to put this logic? 🤔  just thinking

daliclovr (changes requested):

src/Jobs/ChannelMonitorJob.cs
```diff
@@ -0,0 +1,163 @@
...
+        var data = context.JobDetail.JobDataMap;
+        var nodeId = data.GetInt("nodeId");
+        _logger.LogInformation("Starting {JobName} for node {nodeId}... ", nameof(ChannelMonitorJob), nodeId);
+        try
+        {
+            var node1 = await _nodeRepository.GetById(nodeId);
+
+            if (node1 == null)
+            {
+                _logger.LogInformation("The node: {@Node} is no longer ready to be supported quartz jobs", node1);
```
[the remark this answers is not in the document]
RodriFS: I copy-pasted the information message, but it means that there is a job created for a node (maybe just recently) deleted from the NodeGuard. I should change the message.

src/Jobs/ChannelMonitorJob.cs
```diff
@@ -0,0 +1,163 @@
...
+                    if (!addNode.Item1)
+                    {
+                        throw new Exception(addNode.Item2);
+                    }
+                }
+
+                try {
+                    // Recover Operations on channels
+                    await RecoverGhostChannels(node1, node2, channel);
+                    await RecoverChannelInConfirmationPendingStatus(node1);
```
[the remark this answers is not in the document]
RodriFS: good catch, I'll create a try-catch inside the functions instead so they can move on if the previous one fails

src/Services/LightningService.cs
RodriFS: It was there already, I just moved it to it's own function. I agree that this method is insanely large, so we should start thinking about moving stuff

src/Pages/Channels.razor
```diff
@@ -765,7 +765,8 @@
     private bool checkDisableCloseChannelButton(Channel channel)
     {
         ChannelOperationRequest? lastRequest = channel.ChannelOperationRequests.LastOrDefault();
-        if (lastRequest == null && channel.CreatedByNodeGuard == false) return false;
+        if (channel.CreatedByNodeGuard == false) return true;
+        if (lastRequest == null) return false;
```
RodriFS, line 769: Changed this. But I'm not sure why we disable the button

src/Pages/Channels.razor
```diff
@@ -717,7 +717,7 @@
 
     private bool OnWalletFilter(object itemValue, object searchValue)
     {
-        if (itemValue == null) return false;
+        if (itemValue == null) return true;
```
RodriFS, line 720: for now ghost channels will appear with "Unknown" for which wallet opened the channel

src/Jobs/ChannelMonitorJob.cs
```diff
@@ -0,0 +1,152 @@
...
+        {
+            var node1 = await _nodeRepository.GetById(nodeId);
+
+            if (node1 == null)
+            {
+                _logger.LogInformation("The node {NodeId} was set up for monitoring but the node doesn't exist anymore", nodeId);
+                return;
+            }
+
+            var client = LightningService.CreateLightningClient(node1.Endpoint);
```
Jossec101, line 68: Hmm.. creating a gRPC client all the time can lead to resources exhaustion, we should add them via DI

src/Pages/Channels.razor
Jossec101: But, the " ghost" channels were created by NG, we have the info. I don't get this.

src/Pages/Channels.razor
Jossec101: Me neither, worth checking, good point, we shall be able to close channels created from outside.

src/Helpers/Constants.cs
```diff
@@ -55,6 +55,7 @@ public class Constants
 
     // Crons & Jobs
     public static readonly string MONITOR_WITHDRAWALS_CRON = "10 0/5 * * * ?";
+    public static readonly string MONITOR_CHANNELS_CRON = "0 0 0 * * ?";
```
Jossec101, line 58: That's all nights at 00:00AM, shoudn't this be more periodic? like every hour?

src/Pages/Channels.razor
RodriFS: If we have a request in the `ChannelOperationRequest` table, then we know the wallet that originated the funding Tx. But if the channel was originated by lncli for example, or the node had previous channels before it began to be managed by the NodeGuard, then we won't have the info about which wallet created it. But we still would want to manage those channels, right?

src/Jobs/ChannelMonitorJob.cs
RodriFS: Will investigate

src/Pages/Channels.razor
Jossec101: Agree +1

src/Jobs/ChannelMonitorJob.cs
RodriFS: I created a service for reusing clients in jobs.
Other parts of the application (like the lightning service) create the lightning client inside static functions, so we will have to refactor this in order to reuse the client there.

src/Helpers/Constants.cs
Jossec101: Hmm, I would do it more often

RodriFS: moving to every hour

RodriFS:
Added unit testing

src/Jobs/ChannelMonitorJob.cs
Jossec101: Why not this? https://learn.microsoft.com/en-us/aspnet/core/grpc/clientfactory?view=aspnetcore-7.0

Jossec101 (changes requested):
Lets use gRPC client factory by MSFT to have a good way to use gRPC channels efficiently:

https://learn.microsoft.com/en-us/aspnet/core/grpc/clientfactory?view=aspnetcore-7.0

daliclovr (dismissed):

Jossec101 (approved):

PR #257, opened 2023-07-28
Channels opened against nodeguard nodes

RodriFS:
This PR is to show the channels that were opened against nodes managed by nodeguard

Added runners so each person don't have to configure them


.run/Compose Deployment.run.xml
cvicenti, line 1: Hero

cvicenti (approved):

cvicenti (approved):

PR #261, opened 2023-07-31
Lightning client Service

RodriFS:
- Reused the service to get the client channels and now it is the ligthtning client service
- Removed the unmockable dependency


daliclovr:
Does it make sense to pass the client as an argument in `LightningClientServer` if the client cannot be created outside the class?

daliclovr (dismissed):

RodriFS:
> Does it make sense to pass the client as an argument in `LightningClientServer` if the client cannot be created outside the class?

You mean passing the client in the methods as an optional parameter? Even though we're saving the grpc channel in a dictionary, I thought that recreating the lightning client every time for a lot of methods that are going to be called in the same function was a bit of a waste of processing

src/NodeGuard.csproj
```diff
@@ -55,8 +55,6 @@
     <PackageReference Include="Blazorise.SpinKit" Version="1.2.4" />
     <PackageReference Include="RestSharp" Version="106.13.0" />
     <PackageReference Include="Serilog.AspNetCore" Version="7.0.0" />
-    <PackageReference Include="Unmockable" Version="3.0.132" />
-    <PackageReference Include="Unmockable.Wrap" Version="3.0.132" />
```
Jossec101, line 59: ✋

Jossec101 (approved):

PR #267, opened 2023-08-10
Channel Balance reversed for external nodes

RodriFS:


src/Services/LightningService.cs
```diff
@@ -1312,15 +1312,6 @@ await foreach (var response in closeChannelResult.ResponseStream.ReadAllAsync())
...
 
-                    // If the channel is not initiated by a NodeGuard node, we need to swap the balances.
-                    // the balance is always shown from the NodeGuard's perspective
-                    if (!channel.Initiator)
-                    {
-                        localBalance = channel.RemoteBalance + htlcsRemote;
-                        remoteBalance = channel.LocalBalance + htlcsLocal;
-                    }
-
-
```
RodriFS, line 1323: we're always looking through the POV of the managed node, so local balance doesn't have to be reversed

Jossec101 (approved):

PR #270, opened 2023-08-16
Mark closed channel as closed

RodriFS:
- Marking closed channel as closed
- Changed ChannelStatus to ChannelState because there was an enum with that name already, and also it's more appropiate


markettes (approved):
LGTM

PR #273, opened 2023-08-16
Fix channels filter "Opened with".

AlexFront2021:
image

### DIFF
--- a/src/Pages/Channels.razor
+++ b/src/Pages/Channels.razor
@@ -105,6 +105,7 @@
                             {
                                 <SelectItem Value="@(wallet.Id)">@wallet.Name</SelectItem>
                             }
+                              <SelectItem Value="@(_wallets.Count +1)">Unknown</SelectItem>
                         </Select>
                     </FilterTemplate>
                     <DisplayTemplate>
@@ -778,13 +779,24 @@
         return true;
     }
 
-    private bool OnWalletFilter(object? itemValue, object? searchValue)
-    {
-    //If the wallet is null it might be a externally created channel, so we return true
-        if (itemValue == null) return true;
+private bool OnWalletFilter(object? itemValue, object? searchValue)
+{
+    if (searchValue == null)
+        return true;
 
-        return searchValue == null || (int) searchValue == 0 || (int) itemValue == (int) searchValue;
+    if (searchValue is int intValue && (intValue == 0 || intValue == _wallets.Count + 1))
+    {
+        return itemValue == null;
     }
+
+    if (itemValue is int itemIntValue && searchValue is int searchIntValue)
+    {
+        return itemIntValue == searchIntValue;
+    }
+
+    return false;
+}
+
 
     private bool OnSourceNodeIdFilter(object? itemValue, object? searchValue)
     {
--- a/src/Pages/Channels.razor
+++ b/src/Pages/Channels.razor
@@ -105,7 +105,7 @@
                             {
                                 <SelectItem Value="@(wallet.Id)">@wallet.Name</SelectItem>
                             }
-                              <SelectItem Value="@(_wallets.Count +1)">Unknown</SelectItem>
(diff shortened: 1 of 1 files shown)


src/Pages/Channels.razor
```diff
@@ -105,6 +105,7 @@
                             {
                                 <SelectItem Value="@(wallet.Id)">@wallet.Name</SelectItem>
                             }
+                              <SelectItem Value="@(_wallets.Count +1)">Unknown</SelectItem>
```
RodriFS, line 108: I recommend using -1 for this. -1 It's never going a wallet id, so there will never be a mistake

src/Pages/Channels.razor
```diff
@@ -778,14 +779,25 @@
...
+    }
 
-        return searchValue == null || (int) searchValue == 0 || (int) itemValue == (int) searchValue;
+    if (itemValue is int itemIntValue && searchValue is int searchIntValue)
+    {
+        return itemIntValue == searchIntValue;
     }
 
+    return false;
+}
```
RodriFS, line 798: if you're comming back to `All` filter (searchValue == 0), and your item value is a known wallet (not null), then it's not going to appear in your list

RodriFS (changes requested):

src/Pages/Channels.razor
AlexFront2021: With this fix we will get unexpected behaviour 

https://github.com/Elenpay/NodeGuard/assets/81957232/149d0a96-b29b-47bf-a9f2-66906ebce8fe

src/Pages/Channels.razor
AlexFront2021: Anyway the logic will be working because in the case will work the 
`    if (searchValue is int intValue && (intValue == 0 || intValue == _wallets.Count + 1))
    {
        return itemValue == null;
    }`
because itemIntValue == null shows all items.  Your observation is valuable in ensuring the accurate operation of the code and I will add some corrections(but the previous logic will be working as well). Thank you for pointing this out.

RodriFS: if searchValue equals 0 (no filter), and your wallet id is 3, 
your clause will return false, and should return true

src/Pages/Channels.razor
RodriFS:
```suggestion
                              <SelectItem Value= "-1">Unknown</SelectItem>
```

AlexFront2021: Ok. I've fixed according to the comments.

src/Pages/Channels.razor
AlexFront2021: I've added some improvements.

AlexFront2021: or the specific scenario where searchValue is 0 and itemValue is 3, the function would return true, indicating that the item is a match according to the filter criteria.   
```
private bool OnWalletFilter(object? itemValue, object? searchValue)
{
    if (searchValue == null)
        return true;

    if (searchValue is int intValue && (intValue == 0 || intValue == _wallets.Count + 1))
    {
        return itemValue == null;
    }

    if (itemValue is int itemIntValue && searchValue is int searchIntValue)
    {
        return itemIntValue == searchIntValue;
    }

    return false;
}
```

AlexFront2021: We can check via the test 
`
[TestClass]
public class WalletFilterTests
{
    [TestMethod]
    public void TestSearchValueZeroAndItemValueThree()
    {
        // Arrange
        var filter = new YourFilterClass(); // Replace with the actual filter class
        object itemValue = 3;
        object searchValue = 0;

        // Act
        bool result = filter.OnWalletFilter(itemValue, searchValue);

        // Assert
        Assert.IsTrue(result);
    }
}`

RodriFS (approved):

PR #277, opened 2023-08-18
feat: ability to specify the fee rate

markettes:
Fix a problem and added the posibility to specify the fee


src/Rpc/NodeGuardService.cs
```diff
@@ -395,6 +398,11 @@ public override async Task<OpenChannelResponse> OpenChannel(OpenChannelRequest r
                 Changeless = request.Changeless,
             };
 
+            if (request.FeeRate != 0)
+            {
+                channelOperationRequest.FeeRate = request.FeeRate;
+            }
```
RodriFS, line 404: do we need support for economy, fastest, etc?

RodriFS (approved):

src/Rpc/NodeGuardService.cs
Jossec101: I agree, we need to support the types (economy faster and so) but also a custom value and this is mutually exclusive.

Jossec101 (changes requested):

src/Rpc/NodeGuardService.cs
```diff
@@ -358,6 +362,11 @@ public override async Task<OpenChannelResponse> OpenChannel(OpenChannelRequest r
             throw new RpcException(new Status(StatusCode.NotFound, "Wallet not found"));
         }
 
+        if (request.MempoolFeeRate.Equals("") || request.MempoolFeeRate == null)
+        {
```
RodriFS, line 366:
```suggestion
        if (string.IsNullOrEmpty(request.MempoolFeeRate))
        {
```

RodriFS (dismissed):

src/Proto/nodeguard.proto
```diff
@@ -183,4 +199,44 @@ message CloseChannelRequest {
...
+  PENDING = 4;
+  PSBT_SIGNATURES_PENDING = 5;
+  ONCHAIN_CONFIRMATION_PENDING = 6;
+  ONCHAIN_CONFIRMED = 7;
+  FAILED = 8;
+  FINALIZINGPSBT = 9;
+}
+
+enum CHANNEL_OPERATION_TYPE {
+  NULL0 = 0;
```
Jossec101, line 223: NULL0?

src/Proto/nodeguard.proto
```diff
@@ -183,4 +199,44 @@ message CloseChannelRequest {
...
\ No newline at end of file
+}
+
+message GetChannelOperationRequestByIdRequest {
+  // Whether to include requests that are not managed by the NG
+  int32 channel_operation_request_id = 1;
+}
+
+enum CHANNEL_OPERATION_STATUS {
+  NULL = 0;
```
Jossec101, line 210: @markettes We dont need a null = 0 enum, they will never be 0.

Jossec101 (changes requested):

src/Proto/nodeguard.proto
markettes: Null was already taken

src/Proto/nodeguard.proto
markettes: It is just for easing the change from nodeguards enum and also be able to detect not set variables on the request

src/Proto/nodeguard.proto
```diff
@@ -240,3 +243,21 @@ message GetChannelOperationRequestByIdResponse {
   bool private = 12;
   string job_id = 13;
 }
+
+message AddLiquidityRuleRequest {
+// Channel ID as returned by the Lightning implementation
+  int32 channel_id = 1;
```
Jossec101, line 249: I think SCIDs (short channel ids) are `uint64`

Jossec101: https://github.com/lightningnetwork/lnd/blob/d24f12bd0bbc704560c1092a44fc9883a0ef30f8/lnrpc/invoicesrpc/addinvoice.go#L578

src/Rpc/NodeGuardService.cs
```diff
@@ -605,4 +608,145 @@ public override async Task<GetChannelOperationRequestByIdResponse> GetChannelOpe
...
+            throw new RpcException(new Status(StatusCode.NotFound, "Source node not found"));
+        }
+        var destination = await _nodeRepository.GetById(channel.DestinationNodeId);
+        if (destination == null)
+        {
+            throw new RpcException(new Status(StatusCode.NotFound, "Destination node not found"));
+        }
+
+        var node = string.IsNullOrEmpty(source.ChannelAdminMacaroon) ? destination : source;
```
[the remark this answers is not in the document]
markettes: If both nodes are managed this follows the same logic as the liquidity rule logic via the UI so it will keep the consistency

src/Rpc/NodeGuardService.cs
```diff
@@ -605,4 +608,145 @@ public override async Task<GetChannelOperationRequestByIdResponse> GetChannelOpe
...
+        {
+            var addResult = await _liquidityRuleRepository.AddAsync(liquidityRule);
+            if (!addResult.Item1)
+            {
+                throw new RpcException(new Status(StatusCode.Internal, "Error adding liquidity rule"));
+            }
+        }
+        else
+        {
+            liquidityRule.Id = rule.Id;
```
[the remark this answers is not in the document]
markettes: This entity is not yet persisted so I have to add the id of the persisted one to perform the update

src/Proto/nodeguard.proto
markettes: Wrong explanation comment. This is a reference of the ID of the channel in NGs database

Jossec101: NIT: Then maybe you could have edited the `rule` var directly?

src/Proto/nodeguard.proto
```diff
@@ -181,4 +204,60 @@ message CloseChannelRequest {
...
+  optional int32 wallet_id = 8;
+  int32 source_node_id = 9;
+  optional int32 dest_node_id = 10;
+  optional int64 channel_id = 11;
+  bool private = 12;
+  string job_id = 13;
+}
+
+message AddLiquidityRuleRequest {
+// Channel ID as returned by the Lightning implementation
```
Jossec101, line 248:
```suggestion
// Channel ID as in NodeGuard
```

Jossec101 (approved):

PR #280, opened 2023-08-21
Replacing conversion helper for conversion service

RodriFS:
I believe this will give us better error messages for the price converter


cvicenti (approved):

markettes (approved):
LGTM

PR #303, opened 2023-09-13
Import watch-only wallets

Jossec101:
This PR will allow to import segwit singlesig and multisig wallets into NodeGuard, these wallets cannot be used for opening channels rather than as target for swaps and withdrawals destination addresses.


src/Data/Repositories/Interfaces/IWalletRepository.cs
```diff
@@ -65,4 +66,13 @@ public interface IWalletRepository
...
     Task<(bool, string?)> ImportBIP39Wallet(string name, string description, string seedphrase, string derivationPath, string? userId = null);
+
+    /// <summary>
+    /// Imports a watch-only wallet by creating a wallet and its key entities based on the output descriptor
+    /// </summary>
+    /// <param name="name"></param>
+    /// <param name="description"></param>
+    /// <param name="outputDescriptor"></param>
+    /// <returns></returns>
+    Task<(bool,string?)> ImportWatchOnlyWallet(string name, string? description, string outputDescriptor, string? userId = null);
```
daliclovr, line 77: Isn't this automatically set to null if you declare it as `string?`

src/Data/Repositories/Interfaces/IWalletRepository.cs
```diff
@@ -35,8 +35,9 @@ public interface IWalletRepository
     /// <summary>
     /// Obtains all wallets that are Finalised and not in a compromised or archived state
     /// </summary>
+    /// // <param name="includeWatchOnlyWallets">If true, only wallets with an internal wallet will be returned</param>
     /// <returns> List of available wallets</returns>
-    Task<List<Wallet>> GetAvailableWallets();
+    Task<List<Wallet>> GetAvailableWallets(bool includeWatchOnlyWallets = true);
```
daliclovr, line 40: why are you declaring it in the interface as a true by default and then in the respository as false by default?

src/Data/Repositories/WalletRepository.cs
```diff
@@ -385,5 +386,128 @@ public async Task<(bool, string?)> FinaliseWallet(Wallet selectedWalletToFinalis
...
+                        IsFinalised = true,
+                        WalletAddressType = WalletAddressType.NativeSegwit,
+                        IsHotWallet = false,
+                        IsBIP39Imported = false,
+                        ImportedOutputDescriptor = outputDescriptor,
+                        BIP39Seedphrase = null,
+                        Keys = keys,
+                        MofN = multisigDerivationStrategy.RequiredSignatures
+                    };
+                }
```
daliclovr, line 474: Having a factory might be good for these cases so there its less likely to manual errors for some combination of values! No need to refactor now, just a comment :)

src/Data/Repositories/WalletRepository.cs
```diff
@@ -385,5 +386,128 @@ public async Task<(bool, string?)> FinaliseWallet(Wallet selectedWalletToFinalis
...
+
+
+                await TrackAndScanWallet(derivationStrategyBase);
+
+
+            }
+            catch (Exception e)
+            {
+                var errorWhileImportingWatchOnlyWalletFromOutputDescriptor = "Error while importing watch-only wallet from output descriptor";
```
daliclovr, line 505: is there any need for this variable to have such as long name? xD

src/Helpers/WalletParser.cs
```diff
@@ -0,0 +1,108 @@
...
+
+namespace NodeGuard.Helpers;
+
+public static class WalletParser
+{
+    /// <summary>
+    /// Parse the output descriptor string to get the wallet info, Took from BTCPAYServer codebase
+    /// </summary>
+    /// <param name="outputDescriptorStr"></param>
+    public static (DerivationStrategyBase, (BitcoinExtPubKey, RootedKeyPath)[]) ParseOutputDescriptor(
```
daliclovr, line 13: Cannot this be returned in a more structure way than nested tuples?

src/Pages/Channels.razor
```diff
@@ -412,7 +412,7 @@
 
     private async Task FetchData()
     {
-        _availableWallets = await WalletRepository.GetAvailableWallets();
+        _availableWallets = await WalletRepository.GetAvailableWallets(true);
```
daliclovr, line 415: isnt the default false already?

src/Pages/Withdrawals.razor
```diff
@@ -550,7 +550,10 @@
 
         _allWalletRequests = (await WalletWithdrawalRequestRepository.GetAll()).Except(_userPendingRequests).ToList();
 
-        _availableWallets = await WalletRepository.GetAvailableWallets();
+        _availableWallets = await WalletRepository.GetAvailableWallets(false);
+        //TODO Fix BIP39 withdrawals, until then, manual hack to filter them out
```
daliclovr, line 554: create a task for this? :)

src/Services/LightningService.cs
```diff
@@ -740,14 +755,26 @@ public static PSBT GetCombinedPsbt(ChannelOperationRequest channelOperationReque
...
         {
             if (channelOperationRequest == null) throw new ArgumentNullException(nameof(channelOperationRequest));
+            
+            //If the wallet is watch only, we cannot open a channel as there is no way securely sign the PSBT with SIGHASH_NONE (the internal wallet is not there)
+            if (channelOperationRequest.Wallet != null && channelOperationRequest.Wallet.IsWatchOnly)
+            {
+                const string watchOnlyWalletsCannotOpenChannels =
+                    "Watch only wallets cannot open channels";
+
+                _logger?.LogError(watchOnlyWalletsCannotOpenChannels);
```
daliclovr, line 765: maybe logging more info about the channeloperationrequest could be nice, such as an id or even the wallet id which is trying to open the channel

src/Data/Repositories/Interfaces/IWalletRepository.cs
Jossec101: No, this means thats a nullable reference type, it helps linting and null-free dotnet if enforced

src/Data/Repositories/Interfaces/IWalletRepository.cs
Jossec101: Typo, ill fix it

src/Data/Repositories/WalletRepository.cs
Jossec101: Its automatically done by rider 🫶

src/Data/Repositories/Interfaces/IWalletRepository.cs
Jossec101: Fixed

src/Helpers/WalletParser.cs
Jossec101: Tuples are easier to maintain than classes 🤔

src/Pages/ChannelRequests.razor
```diff
@@ -596,7 +596,7 @@
             {
                 _manageableNodes = _manageableNodes.Where(node => node.Id != _selectedDestNode.Id).ToList();
             }
-            _allWallets = await WalletRepository.GetAvailableWallets();
+            _allWallets = await WalletRepository.GetAvailableWallets(false);
```
markettes, line 599: Is it possible to send the funds when closing a channel to a watch-only wallet?

src/Helpers/WalletParser.cs
satinfive: But they are way less readable, when you get the result you need to play with: `result.Item1` and so on which make it not readable enough 🤔  and makes you check the signature of the function to see actually what you are receiving

src/Pages/Nodes.razor
```diff
@@ -208,7 +208,7 @@
     private async Task GetData()
     {
         _nodes = await NodeRepository.GetAllManagedByNodeGuard();
-        _availableWallets = await WalletRepository.GetAvailableWallets();
+        _availableWallets = await WalletRepository.GetAvailableWallets(false);
```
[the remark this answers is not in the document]
satinfive: You make it false by default, you would need to pass true in those cases which makes sense instead of false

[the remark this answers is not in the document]
daliclovr: You already make it false as default, so there is no need to put it in here :)

daliclovr (approved):

src/Pages/Nodes.razor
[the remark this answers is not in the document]
Jossec101: @daliclovr Putting it by design makes me more certain ;), @markettes They could be capable but I don't think they should be used for this reason.

src/Pages/ChannelRequests.razor
Jossec101: Nope, as they are not able to be used for opening channels.

PR #305, opened 2023-09-15
added interceptor to authenticate through api token

daliclovr:
- Added grpc interceptor to handle auth through an api token.
- Addd unittesting for the new interceptor
- Added some logic from past review to UI
- Added creation of APITokens to development environment so services works smoothely in that env.
- Added a feature flag to do the deployment smoothly without downtimes.


src/Data/Repositories/APITokenRepository.cs
```diff
@@ -21,6 +21,12 @@ public class APITokenRepository : IAPITokenRepository
         _dbContextFactory = dbContextFactory;
     }
     
+    private async Task<T> WithContext<T>(Func<ApplicationDbContext, Task<T>> func)
+    {
+        await using var context = _dbContextFactory.CreateDbContext();
+        return await func(context);
+    }
+
```
daliclovr, line 29: @Jossec101 check this. I think it is a fancier method for accessing to the db context instead of calling all the time to this function. You can see usage in `GetByToken` function. I just wanted to try this way and show to the team something different but if it's too much or you don't like it i can just go back to the usual :)

Jossec101: I don't really see the advantage of using callbacks/delegates with generics here, if you ask me what's easier a repeated line to get the context async or a generic method, I will rather go for the 1st one, besides;
```suggestion
    private async Task<T> WithContext<T>(Func<ApplicationDbContext, Task<T>> func)
    {
        await using var context = await _dbContextFactory.CreateDbContextAsync();
        return await func(context);
    }
    
```

Also the code below makes debugging harder if the return is not directly a variable (breakpoints are useless in this case)
:

```return await WithContext(
            async context =>
            {
                var result = await context.ApiTokens.FirstOrDefaultAsync(x => x.TokenHash == token);
                return result;
            }

            );
    }
```

I think this is not KISS, `keep it stupid simple`

test/NodeGuard.Tests/Rpc/AuthInterceptorTests.cs
```diff
@@ -0,0 +1,67 @@
+
+using Grpc.Core;
+using NodeGuard.Data.Models;
+using NodeGuard.Data.Repositories.Interfaces;
+
+namespace NodeGuard.Rpc;
+
+public class AuthInterceptorTests
```
Jossec101, line 8: Can you add Arrange, act, assert comment sections? We've been using them across the whole tests ;)

src/Rpc/AuthInterceptor.cs
```diff
@@ -0,0 +1,36 @@
...
+            throw new RpcException(new Status(StatusCode.Unauthenticated, "No token provided"));
+        }
+
+        var apiToken = await _apiTokenRepository.GetByToken(token);
+        if (apiToken == null)
+        {
+            throw new RpcException(new Status(StatusCode.Unauthenticated, "Invalid token"));
+        }
+
+        return await continuation(request, context);
```
Jossec101, line 33: I always suggest to return a var than directly returning a function, it's easier for debugging later or you will have to evaluate in the IDE manually.

src/Rpc/AuthInterceptor.cs
```diff
@@ -0,0 +1,36 @@
+using Grpc.Core;
+using Grpc.Core.Interceptors;
+using NodeGuard.Data.Repositories.Interfaces;
+
+namespace NodeGuard.Rpc;
+
+public class ServerAuthInterceptor : Interceptor
```
Jossec101, line 7: What about `GRPCAuthInterceptor`? It would be nice also to have what's this class is about using the `///` comments way.

Jossec101: Also the filename should match the class name per conventions.

Jossec101 (changes requested):

src/Rpc/AuthInterceptor.cs
daliclovr: It's true, but it shouldnt be needed to debug this one. Actually in the docs suggested by dotnet documentation is this way since the continuation is a first class object which represent a function and at the is bypassing it... it's not intended to have a result perse

test/NodeGuard.Tests/Rpc/AuthInterceptorTests.cs
daliclovr: fixeeeed

test/NodeGuard.Tests/Rpc/GRPCAuthInterceptorTests.cs
```diff
@@ -0,0 +1,98 @@
...
+        var context = TestServerCallContext.Create();
+        var mockedApiTokenRepository = new Mock<IAPITokenRepository>();
+        var interceptor = new GRPCAuthInterceptor(mockedApiTokenRepository.Object);
+        var continuation = new UnaryServerMethod<string, string>(async (request, context) => { return "response"; });
+
+        var expectedMessage = "No token provided";
+        
+        // Act & Assert
+        var exception = await Assert.ThrowsAsync<RpcException>(() => interceptor.UnaryServerHandler<string, string>(String.Empty, context, continuation));
+        Assert.Equal(expectedMessage, exception.Status.Detail);
```
RodriFS, line 24: You can use "WithMessage" like in this example if, I think it would be more clear what you are asserting

https://github.com/Elenpay/NodeGuard/blob/b52f42a09dad81f87e491fc9117c9d9ac06186a6/test/NodeGuard.Tests/Rpc/NodeGuardServiceTests.cs#L1300

Jossec101: +1, we should use fluent assertions lib, its more easy to read.

Jossec101 (approved):

test/NodeGuard.Tests/Rpc/GRPCAuthInterceptorTests.cs
daliclovr: I am not sure this is "more clear", as the message is actually clear here:
`var expectedMessage = "Invalid token";` 
and having in a string even the inner structure of grpc internal error exceptions in my opinion is verbose since its a well defined structure by the grpc protocol... but in any case guys, i will change it just to follow same structure as the other tests!

daliclovr:
Reviewed. Can someone please review this 🙏

daliclovr:
Tests which are failing are expected to fail since the feature flag is set to false

RodriFS (approved):

PR #307, opened 2023-09-21
New call to retrieve the channel through GRPC

markettes:
feat(nodeguard.proto): add GetChannel RPC method to retrieve channel information by ID

feat(NodeGuardService.cs): implement GetChannel method to retrieve channel information by ID


src/Proto/nodeguard.proto
```diff
@@ -330,3 +335,25 @@ message WithdrawalRequest {
...
 }
+
+message GetChannelRequest {
+  // Channel ID from NGs database
+  int32 channel_id = 1;
+}
+
+enum CHANNEL_STATUS {
+  OPEN = 0;
+  CLOSE = 1;
```
Jossec101, line 346: NIT: Shall be `CLOSED`

Jossec101 (approved):

src/Proto/nodeguard.proto
markettes: Done 🏁

src/Rpc/NodeGuardService.cs
```diff
@@ -859,4 +864,30 @@ public override async Task<GetWithdrawalsRequestStatusResponse> GetWithdrawalsRe
...
+            ChanId = channel.ChanId,
+            SatsAmount = channel.SatsAmount,
+            Status = (CHANNEL_STATUS)((int)channel.Status - 1),
+            CreatedByNodeguard = channel.CreatedByNodeGuard,
+            IsAutomatedLiquidityEnabled = channel.IsAutomatedLiquidityEnabled,
+            IsPrivate = channel.IsPrivate,
+        };
+
+        if (channel.BtcCloseAddress != null)
+            result.BtcCloseAddress = channel.BtcCloseAddress;
```
daliclovr, line 889: Maybe moving this to GetChannelResponse object creation as a ternary operator:

`BtcCloseAddress = channel.btcCloseAddress != null ? channel.BtcCloseAddress : String.Empty` 

this way we avoid having that if :)

src/Rpc/NodeGuardService.cs
```diff
@@ -859,4 +864,30 @@ public override async Task<GetWithdrawalsRequestStatusResponse> GetWithdrawalsRe
...
+            throw new RpcException(new Status(StatusCode.NotFound, "Channel not found"));
+        }
+        
+        var result = new GetChannelResponse()
+        {
+            FundingTx = channel.FundingTx,
+            OutputIndex = channel.FundingTxOutputIndex,
+            ChanId = channel.ChanId,
+            SatsAmount = channel.SatsAmount,
+            Status = (CHANNEL_STATUS)((int)channel.Status - 1),
```
daliclovr, line 882: I would do a direct matching between channel statuses and the possible statuses of `GetChannelResponse`. This line is not readable at all and it miss the point of using enums

Jossec101 (approved):
LGTM

PR #310, opened 2023-10-04
DoS fix when nodes are not reachable over network

Jossec101:
A high amount of exceptions where thrown making the machine unusuable if the nodes are not reachable, a delay of 1s between retries will solve this


markettes (approved):
LGTM

PR #323, opened 2023-10-18
Hotfix: Unable to create wallets

RodriFS:
When we create a new wallet, we check the last part of the derivation path of the last wallet created, and add 1 to generate a wallet with an increased derivation path. We called this the "Internal wallet subderivation path"
With the importing of the descriptor keys, then the last wallets created didn't have an internal wallet subderivation path, so the nodeguard wanted to start from 0 again. We now check only wallets that have an internal wallet finger print, that means it was created by nodeguard


Jossec101 (approved):
Good work ;).

PR #326, opened 2023-10-19
Withdraw all funds fix

RodriFS:
- Added balance check to validator so it warns the user that the balance is not enough in the same modal instead of in a later toast popup
- Removed the Min and Max from the input, so it fails on the validator. Otherwise if you put less than the minimum amount in the input and click save, it changes to the minimum amount but the user doesn't get feedback
- Disabled the coin selector button when "transfer all funds", is selected
- Now when you click "transfer all funds", the input will be filled with the whole balance of the wallet

### DIFF
--- a/src/Pages/Wallets.razor
+++ b/src/Pages/Wallets.razor
@@ -450,7 +450,7 @@
                     <Column ColumnSize="ColumnSize.Is6">
                         <Field>
                             <FieldBody>
-                                <Check TValue="bool" @bind-Checked="@_transferAllFunds">Transfer all funds</Check>
+                                <Check TValue="bool" CheckedChanged="OnTransferAllFundsCheckedChanged" Checked="@_transferAllFunds">Transfer all funds</Check>
                             </FieldBody>
                         </Field>
                     </Column>
@@ -1402,4 +1402,9 @@
         }
     }
 
+    private void OnTransferAllFundsCheckedChanged(bool value)
+    {
+        _transferAllFunds = value;
+        _amountToTransfer = _sourceBalance;
+    }
 }
--- a/src/Pages/Wallets.razor
+++ b/src/Pages/Wallets.razor
@@ -435,7 +435,7 @@
             <Fields>
                 <Validations @ref="_transferFundsValidations">
                     <Column ColumnSize="ColumnSize.Is6">
-                        <Validation Validator="args => ValidationHelper.ValidateWithdrawalAmount(args)">
+                        <Validation Validator="args => ValidationHelper.ValidateWithdrawalAmount(args, _sourceBalance)">
                             <Field>
                                 <FieldLabel>Amount</FieldLabel>
                                 <FieldBody>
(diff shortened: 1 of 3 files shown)


src/Pages/Wallets.razor
```diff
@@ -450,7 +450,7 @@
                     <Column ColumnSize="ColumnSize.Is6">
                         <Field>
                             <FieldBody>
-                                <Check TValue="bool" @bind-Checked="@_transferAllFunds">Transfer all funds</Check>
+                                <Check TValue="bool" CheckedChanged="OnTransferAllFundsCheckedChanged" Checked="@_transferAllFunds">Transfer all funds</Check>
```
markettes, line 453: With this function added `Checked="@_transferAllFunds"` doesn't do anything

src/Pages/Withdrawals.razor
```diff
@@ -160,38 +160,42 @@
...
                             }
                         </DisplayTemplate>
                         <EditTemplate>
-                            <Validation Validator="args => ValidationHelper.ValidateWithdrawalAmount(args)">
-                                <NumericPicker TValue="decimal" @bind-Value="@_amount" Step="0.00001m" CurrencySymbol="₿ " Max="@(_maxWithdrawal)" Min="_minWithdrawal" Decimals="8"  Disabled="@(SelectedUTXOs.Count > 0 || IsCheckedAllFunds)"  />
-                                     <FieldHelp>
+                            <Validation Validator="args => ValidationHelper.ValidateWithdrawalAmount(args, _selectedRequestWalletBalance)">
+                                <NumericPicker TValue="decimal" @bind-Value="@_amount" Step="0.00001m" CurrencySymbol="₿ " Decimals="8" DecimalSeparator="," Disabled="@(SelectedUTXOs.Count > 0 || IsCheckedAllFunds)">
+                                    <Feedback>
+                                        <ValidationError/>
```
markettes, line 173: How does it look the ValidationError plus the FieldHelp? Because I remember weird things happening.

Jossec101 (approved):

src/Pages/Wallets.razor
RodriFS: what do you mean? I'm using the manual event binding as in the docs https://blazorise.com/docs/components/check

src/Pages/Withdrawals.razor
RodriFS: It displaces the FieldHelp down and the error is displayed between the input and the FieldHelp. There is no straightforward way to put it below the FieldHelp because the Feedback tag has to be inside the input

src/Pages/Wallets.razor
markettes: Yes, its just a small thing, that the assignation is being duplicated:
```
private void OnTransferAllFundsCheckedChanged(bool value)
    {
        _transferAllFunds = value;
        _amountToTransfer = _sourceBalance;
    }
```

src/Pages/Withdrawals.razor
markettes: Okay, thats perfect. Just asking because I remembered someone saying something about it. 😄

markettes (approved):

PR #328, opened 2023-10-19
 [Finance manager] Cancel withdrawals.

AlexFront2021:
A finance manager  that does not have the keys of a wallet cannot cancel a withdrawal request.

### DIFF
--- a/src/Pages/Withdrawals.razor
+++ b/src/Pages/Withdrawals.razor
@@ -68,18 +68,16 @@
                     <DataGridColumn TItem="WalletWithdrawalRequest" Filterable="false" Caption="" Sortable="false" Displayable="true">
                         <DisplayTemplate>
                             @*TODO Cancel / Reject  *@
-                            @if ((context.Status == WalletWithdrawalRequestStatus.Pending || context.Status == WalletWithdrawalRequestStatus.PSBTSignaturesPending) && LoggedUser?.Id == context.UserRequestorId)
-
-                            {
-                                <Button Color="Color.Danger" Clicked="()=> ShowRejectCancelModal(context, WalletWithdrawalRequestStatus.Cancelled)">Cancel</Button>
+                            @{
+                                bool canCancel = context.Status == WalletWithdrawalRequestStatus.Pending || context.Status == WalletWithdrawalRequestStatus.PSBTSignaturesPending;
+                                bool isRequestor = LoggedUser?.Id == context.UserRequestorId;
+                                string action = isRequestor ? "Cancel" : "Reject";
+                                WalletWithdrawalRequestStatus newStatus = isRequestor ? WalletWithdrawalRequestStatus.Cancelled : WalletWithdrawalRequestStatus.Rejected;
                             }
-                            else if ((context.Status == WalletWithdrawalRequestStatus.Pending || context.Status == WalletWithdrawalRequestStatus.PSBTSignaturesPending)
-                                     && LoggedUser?.Id != context.UserRequestorId)
+                            
+                            @if (canCancel || (canCancel && !isRequestor))
                             {
-                                <Button Color="Color.Danger" Clicked="()=> ShowRejectCancelModal(context, WalletWithdrawalRequestStatus.Rejected)">Reject</Button>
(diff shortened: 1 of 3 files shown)


src/Pages/Withdrawals.razor
```diff
@@ -77,6 +77,9 @@
                                      && LoggedUser?.Id != context.UserRequestorId)
                             {
                                 <Button Color="Color.Danger" Clicked="()=> ShowRejectCancelModal(context, WalletWithdrawalRequestStatus.Rejected)">Reject</Button>
+                            } else if (LoggedUser?.Id == context.UserRequestorId)
+                            {
+                                <Button Color="Color.Danger" Clicked="()=> ShowRejectCancelModal(context, WalletWithdrawalRequestStatus.Cancelled)">Cancel</Button>
```
RodriFS, line 82: You really don't need this 3rd clause, it will never run.

I think you can even simplify it to
```
@if (LoggedUser?.Id == context.UserRequestorId)
{
    <Button Color="Color.Danger" Clicked="()=> ShowRejectCancelModal(context, WalletWithdrawalRequestStatus.Cancelled)">Cancel</Button>
}
else 
{
    <Button Color="Color.Danger" Clicked="()=> ShowRejectCancelModal(context, WalletWithdrawalRequestStatus.Rejected)">Reject</Button>
}
```

src/Pages/Withdrawals.razor
```diff
@@ -521,8 +524,13 @@
...
     private async Task GetData()
     {
-        if (LoggedUser?.Id != null)
+        if (LoggedUser?.Id != null && !_isFinanceManager)
+        {
             _userPendingRequests = await WalletWithdrawalRequestRepository.GetUnsignedPendingRequestsByUser(LoggedUser.Id);
+        }else 
+        {
+            _userPendingRequests = await WalletWithdrawalRequestRepository.GetAllUnsignedPendingRequests(); 
+        }
```
RodriFS, line 533: you don't need the first function anymore, because now it doesn't matter who the user is when fetching, only when displaying the button. Remember that the table above doesn't show if the user is not a finance manager. So you can simplify all this by


```
if (LoggedUser?.Id != null && _isFinanceManager)
{
   _userPendingRequests = await WalletWithdrawalRequestRepository.GetAllUnsignedPendingRequests(); 
}
```

Check if `GetUnsignedPendingRequestsByUser` is used elsewhere, otherwise you can remove it entirely from the code

RodriFS (changes requested):
Hey @AlexFront2021, Your code works, but I'm requesting some improvements.

Nice job!

src/Pages/Withdrawals.razor
AlexFront2021: Yes, it make sense.

src/Pages/Withdrawals.razor
```diff
@@ -68,15 +68,16 @@
...
+                            @{
+                                bool canCancel = context.Status == WalletWithdrawalRequestStatus.Pending || context.Status == WalletWithdrawalRequestStatus.PSBTSignaturesPending;
+                                bool isRequestor = LoggedUser?.Id == context.UserRequestorId;
+                                string action = isRequestor ? "Cancel" : "Reject";
+                                WalletWithdrawalRequestStatus newStatus = isRequestor ? WalletWithdrawalRequestStatus.Cancelled : WalletWithdrawalRequestStatus.Rejected;
                             }
-                            else if ((context.Status == WalletWithdrawalRequestStatus.Pending || context.Status == WalletWithdrawalRequestStatus.PSBTSignaturesPending)
-                                     && LoggedUser?.Id != context.UserRequestorId)
+                            
+                            @if (canCancel || (canCancel && !isRequestor))
```
RodriFS, line 78: This codes works like this:
if canCancel is true, the first part is true, the code displays the button
if canCancel is false, first part is false, the second part is also false, the code doesn't display the button

so you only need the first part:
```
@if (canCancel) {
```

AlexFront2021: Sure, I missed.

RodriFS (approved):

PR #330, opened 2023-10-20
Docker compose and visual fix

RodriFS:
- Changed nbxplorer version to elenpay-develop in docker compose file
- Added a visual fix so the row doen't expand when you click the dropdown button

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -17,7 +17,7 @@ services:
 
   nbxplorer:
     restart: unless-stopped
-    image: ghcr.io/elenpay/nbxplorer:v2.3.65-1.0.0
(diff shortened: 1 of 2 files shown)


docker/docker-compose.yml
```diff
@@ -17,7 +17,7 @@ services:
 
   nbxplorer:
     restart: unless-stopped
-    image: ghcr.io/elenpay/nbxplorer:v2.3.65-1.0.0
+    image: ghcr.io/elenpay/nbxplorer:elenpay-develop
```
daliclovr, line 20: why are we doing this?

Jossec101: Pointing to a fixed version is wrong, we lost features with this typo.

Jossec101 (approved):

PR #331, opened 2023-10-25
feat: Bitcoin Output descriptors for wallets

markettes:


test/NodeGuard.Tests/Helpers/WalletParserTests.cs
```diff
@@ -149,4 +150,315 @@ public void ParseOutputDescriptor_InvalidSegwitSingleSig(string descriptor, stri
 
     }
     
+    // Testing GetOutputDescriptor method
+    [Fact]
+    public void GetOutputDescriptor_NativeSegwits()
```
RodriFS, line 155: Please if you can order the tests as Arrange/Act/Assert

RodriFS (dismissed):

test/NodeGuard.Tests/Helpers/WalletParserTests.cs
markettes: Done! 🏁

src/Helpers/WalletParser.cs
```diff
@@ -105,4 +108,99 @@ DerivationStrategyBase Parse(string str)
             return strategy;
         }
     }
+
+    public static string GetOutputDescriptor(Wallet wallet, string bitcoinNetwork)
```
Jossec101, line 112: NIT: We could use https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods here

src/Helpers/WalletParser.cs
```diff
@@ -105,4 +108,99 @@ DerivationStrategyBase Parse(string str)
...
+    }
+    
+    public static byte[] GetMasterFingerprint(string masterFingerprint)
+    {
+        var internalBytes = Enumerable.Range(0, masterFingerprint.Length)
+            .Where(x => x % 2 == 0)
+            .Select(x => Convert.ToByte(masterFingerprint.Substring(x, 2), 16))
+            .ToArray();
+        return internalBytes;
+    }
```
Jossec101, line 205: Can you document this function?

Jossec101 (changes requested):
What about checksums? We should enforce them.

src/Helpers/WalletParser.cs
markettes: done 🏁

Jossec101 (approved):

PR #339, opened 2023-11-02
Settlement view in NG

AleksSoftware:
- Add the username of the BTCPayServer’s user under the “Requestor” field
- In the description change “BTCPayServer” with “ElenPAY’s backoffice” and instead of displaying the storeID, let’s show the store name
- For automatic settlements let’s put “Auto”
- For manual settlements let’s use the username before the @
The PR connected with PR btcpayserver https://github.com/Elenpay/btcpayserver/pull/96

### DIFF
--- a/src/Pages/Withdrawals.razor
+++ b/src/Pages/Withdrawals.razor
@@ -7,16 +7,17 @@
 @using NodeGuard.Jobs
 @using Google.Protobuf
 @using NBXplorer.Models
+@using Newtonsoft.Json
 @using NodeGuard.Helpers;
 @using Polly
 @attribute [Authorize(Roles = "Superadmin,NodeManager,FinanceManager")]
 
 <Row>
-    @if (_isFinanceManager)
-    {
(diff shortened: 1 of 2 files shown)


src/Pages/Withdrawals.razor
```diff
@@ -581,11 +579,36 @@
...
             _userPendingRequests.Remove(arg.Item);
-
         }
+    }
+
+private string ShowRequestor(WalletWithdrawalRequest walletWithdrawalRequest)
+{
+    if (walletWithdrawalRequest.RequestMetadata != null)
+    {
+        var requestMetadata = JsonConvert.DeserializeObject<Dictionary<string, string>>(walletWithdrawalRequest.RequestMetadata);
```
RodriFS, line 589: What happens if the request metadata is not a json string, for example a plain text?

AleksSoftware: Make sense to wrap try-catch block. Btw we can manage the type of metadata.

RodriFS (approved):

PR #342, opened 2023-11-09
Feature to add swaps to an address instead of a wallet

markettes:
Wait for the merge, since Liquidator and other components depend on this.

### DIFF
--- a/src/Data/ApplicationDbContext.cs
+++ b/src/Data/ApplicationDbContext.cs
@@ -43,13 +43,21 @@ namespace NodeGuard.Data
         protected override void OnModelCreating(ModelBuilder modelBuilder)
         {
             modelBuilder.Entity<ChannelOperationRequest>()
-                    .HasOne(cor => cor.SourceNode)
-                    .WithMany(node => node.ChannelOperationRequestsAsSource)
-                    .HasForeignKey(cor => cor.SourceNodeId);
+                .HasOne(cor => cor.SourceNode)
+                .WithMany(node => node.ChannelOperationRequestsAsSource)
+                .HasForeignKey(cor => cor.SourceNodeId);
             modelBuilder.Entity<ChannelOperationRequest>()
                 .HasOne(cor => cor.DestNode)
                 .WithMany(node => node.ChannelOperationRequestsAsDestination)
                 .HasForeignKey(cor => cor.DestNodeId);
+            modelBuilder.Entity<LiquidityRule>()
+                .HasOne(lr => lr.SwapWallet)
+                .WithMany(wallet => wallet.LiquidityRulesAsSwapWallet)
+                .HasForeignKey(lr => lr.SwapWalletId);
+            modelBuilder.Entity<LiquidityRule>()
+                .HasOne(lr => lr.ReverseSwapWallet)
+                .WithMany(wallet => wallet.LiquidityRulesAsReverseSwapWallet)
+                .HasForeignKey(lr => lr.ReverseSwapWalletId);
 
             modelBuilder.Entity<Node>().HasIndex(x => x.PubKey).IsUnique();
             modelBuilder.Entity<Wallet>().HasIndex(x => new {x.InternalWalletSubDerivationPath, x.InternalWalletMasterFingerprint}).IsUnique();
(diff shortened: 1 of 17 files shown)


src/Rpc/NodeGuardService.cs
```diff
@@ -693,18 +694,47 @@ public override async Task<AddLiquidityRuleResponse> AddLiquidityRule(AddLiquidi
...
+        var swapWallet = await _walletRepository.GetById(request.SwapWalletId);
+        if (swapWallet == null)
         {
-            throw new RpcException(new Status(StatusCode.NotFound, "Wallet not found"));
+            throw new RpcException(new Status(StatusCode.NotFound, "Swap wallet not found"));
+        }
+
+        if (request.IsReverseSwapWalletRule)
+        {
+            if (!request.HasReverseSwapWalletId)
```
RodriFS, line 705: why is this necessary? maybe is better to compare `request.ReverseSwapWalletId == null` and have one less field that might not be updated

src/Rpc/NodeGuardService.cs
```diff
@@ -693,18 +694,47 @@ public override async Task<AddLiquidityRuleResponse> AddLiquidityRule(AddLiquidi
...
+                throw new RpcException(new Status(StatusCode.FailedPrecondition, "WalletId is required for wallet rules"));
+            
+            var wallet = await _walletRepository.GetById(request.ReverseSwapWalletId);
+            if (wallet == null)
+                throw new RpcException(new Status(StatusCode.NotFound, "Wallet not found"));
+        }
+        else
+        {
+            if (!request.HasReverseSwapAddress)
```
RodriFS, line 714: same here

RodriFS (approved):

src/Rpc/NodeGuardService.cs
markettes: This is more of a double check, if someone forgot to remove when changing and both fields were populated this field would be essential to determine the way to execute the swap.

RodriFS: I meant that the field seems unnecessary, not the check. But is just to be nitpicky

src/Data/ApplicationDbContext.cs
```diff
@@ -43,13 +43,21 @@ protected override void OnConfiguring(DbContextOptionsBuilder optionsBuilder)
...
                 .WithMany(node => node.ChannelOperationRequestsAsDestination)
                 .HasForeignKey(cor => cor.DestNodeId);
+            modelBuilder.Entity<LiquidityRule>()
+                .HasOne(lr => lr.SwapWallet)
+                .WithMany(wallet => wallet.LiquidityRulesAsSwapWallet)
+                .HasForeignKey(lr => lr.SwapWalletId);
+            modelBuilder.Entity<LiquidityRule>()
+                .HasOne(lr => lr.ReverseSwapWallet)
+                .WithMany(wallet => wallet.LiquidityRulesAsReverseSwapWallet)
+                .HasForeignKey(lr => lr.ReverseSwapWalletId);
```
Jossec101, line 60: NIT: Using the fluent api for one-to-many is not strictly needed.

src/Pages/Channels.razor
```diff
@@ -303,23 +303,66 @@
...
+                                    <Validation Validator="ValidationRule.IsSelected">
+                                        <FieldLabel>Wallet to use in SWAPS operations</FieldLabel>
+                                        <SelectList TItem="Wallet"
+                                                    TValue="int"
+                                                    Data="@_availableWallets"
+                                                    SelectedValue="_currentLiquidityRule.SwapWalletId"
+                                                    TextField="@((item) => item.Name)"
+                                                    ValueField="@((item) => item.Id)"
+                                                    SelectedValueChanged="@OnSelectedSwapWalletChanged"
+                                                    DefaultItemText="Choose the wallet to fund/receive swap amounts">
```
Jossec101, line 319:
```suggestion
                                                    DefaultItemText="Choose the wallet to fund or receive the swaps">
```

Jossec101 (approved):

src/Data/ApplicationDbContext.cs
markettes: It was asking me for it

RodriFS (approved):

PR #344, opened 2023-11-20
Node disable/enabled.

AleksSoftware:
- Convert the action buttons to a dropdown menu to keep consistency
- Add a `Disable Node/Enable Node` button in dropdown
- On node deletion, instead of deleting the node, remove the endpoint and macaroon. Make sure the node is not shown in the managed view anymore
- On node creation, make sure that if you re-add the same node (Same pubkey), it reuses the previously created one.


src/Pages/Nodes.razor
```diff
@@ -237,7 +262,19 @@
...
             arg.Item.CreationDatetime = DateTimeOffset.Now;
-            var addResult = await NodeRepository.AddAsync(arg.Item);
+            (bool, string?) addResult;
+            var restoredNode = await NodeRepository.GetByPubkey(arg.Item.PubKey);
+            if (restoredNode != null)
+            {
+                restoredNode.EnabledNode = true;
+                restoredNode.Endpoint = arg.Item.Endpoint;
+                restoredNode.ChannelAdminMacaroon = arg.Item.ChannelAdminMacaroon;
+                addResult=  NodeRepository.Update(restoredNode);
```
RodriFS, line 272: What if the user changed the other properties like name and description, etc?
You should update all fields of restoredNode

RodriFS (commented):
Remember to make sure that the Jobs are not called anymore when you disable a node, and that jobs scheduled to run are cancelled and deleted

src/Pages/Nodes.razor
```diff
@@ -326,6 +370,21 @@
...
 
     }
+    
+    private async Task StopJob(Node node)
+    {
+        try
+        {
+            IScheduler scheduler = await SchedulerFactory.GetScheduler();
+
+            await scheduler.UnscheduleJob(new TriggerKey(node.JobId));
```
RodriFS, line 380: does this only uneschedule the job or also deletes the job from the database?

src/Pages/Nodes.razor
```diff
@@ -205,9 +216,27 @@
...
+       return node.EnabledNode ? "Disable Node" : "Enable Node";
+    }
+
+    private async Task DisabledNode(Node node)
+    {
+        if (node.EnabledNode)
+        {
+            await StopJob(node);
+        }
+        node.EnabledNode = !node.EnabledNode;
```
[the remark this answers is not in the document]
RodriFS: Also, what happens when you re-enable the node? you should schedule the jobs again

RodriFS (changes requested):

test/NodeGuard.Tests/Rpc/NodeGuardServiceTests.cs
```diff
@@ -1104,7 +1104,7 @@ public async Task GetNodes_RequestIncludeUnmanaged_ReturnsUnmanagedNodes()
             };
 
             nodeRepositoryMock.Setup(repo => repo.GetAll()).ReturnsAsync(sampleNodes);
-            nodeRepositoryMock.Setup(repo => repo.GetAllManagedByNodeGuard()).ReturnsAsync(new List<Node>());
+            nodeRepositoryMock.Setup(repo => repo.GetAllManagedByNodeGuard(true)).ReturnsAsync(new List<Node>());
```
RodriFS, line 1107: You don't need to modify the tests, because you already added a default in the parameter
```
public Task<List<Node>> GetAllManagedByNodeGuard(bool includeDisabledNodes = true)
```

src/Pages/ChannelRequests.razor
```diff
@@ -80,7 +80,7 @@
                                 </TextEdit>
                             </Addon>
                             <Addon AddonType="AddonType.End">
-                                <Button Color="Color.Primary" Clicked="@SearchNode">Search</Button>
+                                <Button Color="Color.Primary" Clicked="@(() => SearchNode())">Search</Button>
```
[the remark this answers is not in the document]
markettes: We were doing some debugging and this change allowed me to debug it. But checking again now this change has no effect.

src/Data/Repositories/NodeRepository.cs
```diff
@@ -109,19 +110,26 @@ public async Task<List<Node>> GetAll()
...
                 .ThenInclude(x => x.Keys)
-                .Where(node => node.Endpoint != null)
-                .ToListAsync();
+                .Include(x => x.ReturningFundsWallet)
+                .Where(node => node.Endpoint != null);
+
+            if (includeDisabledNodes)
+            {
+                query = query.Where(node => node.EnabledNode);
+            }
```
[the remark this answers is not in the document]
markettes: I think it should be like this instead
```
if (!includeDisabledNodes)
{
    query = query.Where(node => node.EnabledNode);
}
```

src/Pages/Nodes.razor
```diff
@@ -27,16 +27,25 @@
...
+                            <DropdownToggle Color="Color.Primary">
+                            </DropdownToggle>
+                            <DropdownMenu>
+                                <DropdownItem Clicked="context.Clicked">Edit</DropdownItem>
+                                <DropdownItem Clicked="@(() => ShowDeleteDialog(context.Item))">Delete</DropdownItem>
+                                <DropdownItem Clicked="@(() => DisabledNode(context.Item) )">@_NodesState(context.Item)</DropdownItem>
+                            </DropdownMenu>
+                        </Dropdown>
+                    </EditCommandTemplate>
+                    <DeleteCommandTemplate/>
```
markettes, line 44: This line should be deleted right?

src/Pages/Nodes.razor
```diff
@@ -102,6 +111,7 @@
 
                     </EditTemplate>
                 </DataGridColumn>
+                <DataGridColumn TItem="Node" Filterable="false" Editable="true" Field="@nameof(Node.EnabledNode)" Caption="Enabled Node" Sortable="false" Displayable="@IsColumnVisible(NodesColumnName.EnabledNode)"></DataGridColumn>
```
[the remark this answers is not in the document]
markettes: If this line is editable this attribute could be changed from the edit node pop-up. Do we want that?

src/Pages/Nodes.razor
```diff
@@ -27,16 +27,25 @@
...
+                <DataGridCommandColumn TItem="Node" Filterable="false"  Editable="false" Width="60px">
+
+                    <EditCommandTemplate>
+                        <Dropdown @onclick:stopPropagation="true">
+                            <DropdownToggle Color="Color.Primary">
+                            </DropdownToggle>
+                            <DropdownMenu>
+                                <DropdownItem Clicked="context.Clicked">Edit</DropdownItem>
+                                <DropdownItem Clicked="@(() => ShowDeleteDialog(context.Item))">Delete</DropdownItem>
+                                <DropdownItem Clicked="@(() => DisabledNode(context.Item) )">@_NodesState(context.Item)</DropdownItem>
```
markettes, line 40:
```suggestion
                                <DropdownItem Clicked="@(() => DisabledNode(context.Item))">@_NodesState(context.Item)</DropdownItem>
```

src/Pages/Nodes.razor
```diff
@@ -308,6 +351,7 @@
         {
             if (node != null)
             {
+                await StopJob(node);
                 var (result, _) = NodeRepository.Remove(node);
                 if (!result)
```
markettes, line 356: If the node cannot be removed, shouldn't the job be restored?

test/NodeGuard.Tests/Rpc/NodeGuardServiceTests.cs
markettes: It would be nice to create a test case for this. I know is a simple thing, but it's nice to have.

src/Pages/Nodes.razor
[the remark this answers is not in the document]
markettes: I think it should be like this instead
```
if (!includeDisabledNodes)
{
    query = query.Where(node => node.EnabledNode);
}
```

src/Pages/Nodes.razor
AleksSoftware: just uneschedule right, make sense to delete from db.

src/Pages/Nodes.razor
markettes: No, I mean the delete command template tag. Because there is already a delete button up in the code.

src/Pages/Nodes.razor
markettes: But if the user doesn't get to delete the node it leaves a state where a node is in the system but the jobs for it are not running. It doesn't sound good to me.

src/Pages/Nodes.razor
```diff
@@ -235,9 +264,23 @@
         }
         else
         {
-            arg.Item.UpdateDatetime = DateTimeOffset.Now;
-            arg.Item.CreationDatetime = DateTimeOffset.Now;
-            var addResult = await NodeRepository.AddAsync(arg.Item);
+            Node newNode = arg.Item as Node;
+            newNode.UpdateDatetime = DateTimeOffset.Now;
+            newNode.CreationDatetime = DateTimeOffset.Now;
```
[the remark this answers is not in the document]
AleksSoftware: Yes, I understand that I'm repeating myself, but on line 275, I reassigned the object newNode to restoredNode. The UpdateDatetime values still contain the old values from the restored node. To address this, I reassigned it again to refresh the UpdateDatetime.

src/Pages/Nodes.razor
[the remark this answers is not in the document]
markettes: Should be this attribute editable? Do we want it to be edited from the edit pop-up?

src/Data/Repositories/NodeRepository.cs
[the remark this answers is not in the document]
markettes: I think it should be like this 
```
if (!includeDisabledNodes)
{
    query = query.Where(node => node.EnabledNode);
}
```

src/Pages/Nodes.razor
AleksSoftware: So we need to stop the job before deletion the node.

src/Data/Repositories/NodeRepository.cs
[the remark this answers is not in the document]
RodriFS: yes, you're absolutely right

src/Pages/Nodes.razor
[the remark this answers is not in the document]
RodriFS: true, editable should be false

src/Pages/Nodes.razor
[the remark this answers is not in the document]
RodriFS: if it is an existing node, the `Update` function will take care of updating the `UpdateDatetime` everytime you call it. And your line in 277 will take care of the `CreationDatetime`
if is a new node, `UpdateDatetime` should be null and the `AddSync` function will take care of the `CreationDatetime`

test/NodeGuard.Tests/Rpc/NodeGuardServiceTests.cs
RodriFS: might be because of the mock. Leave it like it was

src/Pages/Nodes.razor
AleksSoftware: In this case we will have an extra delete button 
image

src/Data/Repositories/NodeRepository.cs
```diff
@@ -109,19 +110,21 @@ public async Task<List<Node>> GetAll()
...
 
-            var resultAsync = await applicationDbContext.Nodes
+            var query = applicationDbContext.Nodes
                 .Include(x => x.ReturningFundsWallet)
                 .ThenInclude(x => x.Keys)
-                .Where(node => node.Endpoint != null)
-                .ToListAsync();
+                .Include(x => x.ReturningFundsWallet)
+                .Where(node => node.Endpoint != null);
+
```
markettes, line 122: Is it missing the conditional for including/excluding the disabled nodes?

AleksSoftware: Now now sense to add this the logic because if node disabled anyway it should be visible.

markettes: But then why having the parameter in the function?

RodriFS: The parameter is not needed anymore. Enable/disabled nodes should be always visible and you can see the status in a new column. Soft deleted nodes are the ones that shouldn't be visible and that is taken care of by deleting the endpoint and macaroon

src/Helpers/JobTypes.cs
```diff
@@ -44,6 +46,48 @@ public class SimpleJob
...
         return new JobAndTrigger(job, trigger);
     }
+    
+    public static async Task Reschedule<T>(IScheduler scheduler, string identitySuffix) where T : IJob
+    {
+        try
+        {
+            var triggerKey = new TriggerKey($"{typeof(T).Name}-{identitySuffix}");
+            
+            await scheduler.UnscheduleJob(triggerKey);
```
RodriFS, line 56: do you really need this part?

src/Helpers/JobTypes.cs
```diff
@@ -103,7 +148,9 @@ public static async Task Execute(IJobExecutionContext context, Func<Task> callba
         if (intervals == null)
         {
             throw new Exception("No interval list found, make sure you're using the RetriableJob class");
-        };
+        }
+
+        ;
```
RodriFS, line 153: there's something wrong with your formatter

src/Helpers/JobTypes.cs
```diff
@@ -134,7 +181,9 @@ public static async Task OnFail(IJobExecutionContext context, Func<Task> callbac
...
```
RodriFS, line 186: same here

src/Helpers/JobTypes.cs
```diff
@@ -156,7 +205,9 @@ public static async Task OnFail(IJobExecutionContext context, Func<Task> callbac
...
```
RodriFS, line 210: same here

src/Migrations/20231126115717_DisableNode.cs
```diff
@@ -0,0 +1,22 @@
...
+
+        }
+
+        /// <inheritdoc />
+        protected override void Down(MigrationBuilder migrationBuilder)
+        {
+
+        }
+    }
+}
```
RodriFS, line 22: empty migration??

src/Migrations/20231126115717_DisableNode.Designer.cs
```diff
@@ -0,0 +1,1263 @@
+﻿// <auto-generated />
```
RodriFS, line 1: This file shouldn't exist

src/Migrations/20231127171339_RenamedDisableNode.Designer.cs
```diff
@@ -0,0 +1,1263 @@
+﻿// <auto-generated />
```
RodriFS, line 1: There should only be 1 designer file, for creating IsNodeEnabled. Please remove all of the designer and migrations you created on this branch, drop the database, and redo the migration

src/Migrations/20231127171339_RenamedDisableNode.cs
```diff
@@ -0,0 +1,28 @@
...
+        }
+
+        /// <inheritdoc />
+        protected override void Down(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.RenameColumn(
+                name: "IsNodeEnabled",
+                table: "Nodes",
+                newName: "EnabledNode");
+        }
```
RodriFS, line 26: this will fail. read comment above

src/Pages/Nodes.razor
```diff
@@ -205,9 +215,48 @@
...
+          
+            if (node.IsNodeEnabled)
+            {
+                await StopJob(node);
+            }
+            else
+            {
+                await CreateJobs(node);
+            }
+            node.IsNodeEnabled = !node.IsNodeEnabled;
```
RodriFS, line 236: this is confusing. Please first update the state of the boolean and then do the if statement

src/Pages/Nodes.razor
```diff
@@ -205,9 +215,48 @@
...
+            node.JobId = null;
+            _NodesState(node);
+            var (result, _) = NodeRepository.Update(node);
+            _nodes = await NodeRepository.GetAllManagedByNodeGuard(false);
+            _availableWallets = await WalletRepository.GetAvailableWallets();
```
RodriFS, line 241: this should only be done if result was successful

src/Pages/Nodes.razor
```diff
@@ -235,9 +284,23 @@
...
+            var restoredNode = await NodeRepository.GetByPubkey(arg.Item.PubKey);
+         
+            if (restoredNode != null)
+            {
+                newNode = restoredNode;
+                newNode.Endpoint = arg.Item.Endpoint;
+                newNode.PubKey = arg.Item.PubKey;
+                newNode.ChannelAdminMacaroon = arg.Item.ChannelAdminMacaroon;
+                newNode.IsNodeEnabled = true;
+                addResult =  NodeRepository.Update(newNode);
```
RodriFS, line 298: what about name, description, returning funds wallet, autosweep?

src/Pages/Nodes.razor
```diff
@@ -319,36 +385,71 @@
                     _nodes = await NodeRepository.GetAll();
 
                 }
-
+                await StopJob(node);
```
RodriFS, line 388: I think you should delete the jobs, not just unschedule them. @Jossec101?

src/Pages/Nodes.razor
```diff
@@ -205,9 +215,48 @@
...
+            {
+                await StopJob(node);
+            }
+            else
+            {
+                await CreateJobs(node);
+            }
+            node.IsNodeEnabled = !node.IsNodeEnabled;
+            node.JobId = null;
+            _NodesState(node);
```
RodriFS, line 238: this doesn't do anything, since it doesn't modify the node in place

src/Pages/Nodes.razor
```diff
@@ -319,36 +385,71 @@
...
+    {
+        try
+        {
+            IScheduler scheduler = await SchedulerFactory.GetScheduler();
+            
+            await SimpleJob.UnscheduleJob<NodeChannelSuscribeJob>(scheduler, node.Id.ToString());
+            await SimpleJob.UnscheduleJob<ProcessNodeChannelAcceptorJob>(scheduler, node.Id.ToString());
+            
+            node.JobId = null;
+            _NodesState(node);
```
RodriFS, line 407: same

src/Pages/Nodes.razor
```diff
@@ -319,36 +385,71 @@
...
             IScheduler scheduler = await SchedulerFactory.GetScheduler();
 
             var map = new JobDataMap();
-            map.Put("nodeId", node.Id);
-
-            var job = SimpleJob.Create<NodeChannelSuscribeJob>(map, node.Id.ToString());
-            await scheduler.ScheduleJob(job.Job, job.Trigger);
-
-            node.JobId = job.Job.Key.ToString();
+            var isJobExists = await SimpleJob.IsJobExists<NodeChannelSuscribeJob>(scheduler, node.Id.ToString());
```
RodriFS, line 422: Maybe you can create a function `Upsert` instead to avoid so much repetition?

src/Pages/Nodes.razor
AleksSoftware: 🤔 While this might not be overly confusing, I personally don't find it to be a significant issue.

src/Pages/Nodes.razor
AleksSoftware: They were assigned from restoreNode.

src/Pages/Nodes.razor
AleksSoftware: Now no sense for the value.

src/Pages/Nodes.razor
AleksSoftware: Ok, could be improved (DRY)

src/Pages/Nodes.razor
AleksSoftware: The last time we we discussed about unschedule?

src/Pages/Nodes.razor
```diff
@@ -235,9 +284,23 @@
...
-            var addResult = await NodeRepository.AddAsync(arg.Item);
+            Node newNode = arg.Item as Node;
+            (bool, string?) addResult;
+            var restoredNode = await NodeRepository.GetByPubkey(arg.Item.PubKey);
+         
+            if (restoredNode != null)
+            {
+                newNode = restoredNode;
+                newNode.Endpoint = arg.Item.Endpoint;
+                newNode.PubKey = arg.Item.PubKey;
```
RodriFS, line 295: pubkey is not needed, it should be the same

src/Pages/Nodes.razor
RodriFS: what if the person renamed the node, or activated autosweep when they re-added the node?

src/Pages/Nodes.razor
RodriFS: we talked about this:
New node:
- Create Job
Deleting the node:
- interrupt and delete job
Disable node:
- interrupt and unschedule the job
Enable node:
- re-schedule job

src/Pages/Nodes.razor
AleksSoftware: Fixed as we discussed.

src/Helpers/JobTypes.cs
AleksSoftware: Removed

src/Pages/Nodes.razor
```diff
@@ -235,9 +282,28 @@
...
+            if (restoredNode != null)
+            {
+                newNode = restoredNode;
+                newNode.Endpoint = arg.Item.Endpoint;
+                newNode.ChannelAdminMacaroon = arg.Item.ChannelAdminMacaroon;
+                newNode.IsNodeEnabled = true;
+                newNode.Description =  arg.Item.Description;
+                newNode.AutosweepEnabled = arg.Item.AutosweepEnabled;
+                newNode.Name = arg.Item.Name;
+                newNode.ReturningFundsWalletId = arg.Item.ReturningFundsWalletId;
```
RodriFS, line 298: have you tried adding the node id to the new item and calling update? Maybe it works
```
Node newNode = arg.Item as Node;
newNode.Id = restoredNode.Id
addResult = NodeRepository.Update(newNode);
```

src/Pages/Nodes.razor
```diff
@@ -204,6 +214,43 @@
...
+    }
+    
+    private async Task ToggleNodeStatus(Node node)
+    {
+        try
+        {
+            node.IsNodeEnabled = !node.IsNodeEnabled;
+            if (!node.IsNodeEnabled)
+            {
+                await StopJob(node);
```
RodriFS, line 230: this deletes the job, it should unschedule the job.
Disabling node: uneschedule job
Deleting node: delete job

RodriFS (changes requested):

src/Pages/Nodes.razor
AleksSoftware: Yes, it works.

src/Pages/Nodes.razor
AleksSoftware: As we discuss before to delete the job.

src/Pages/Nodes.razor
AleksSoftware: Ok, it could be improved.

src/Pages/Nodes.razor
```diff
@@ -319,41 +382,77 @@
...
-            var acceptorJob = SimpleJob.Create<ProcessNodeChannelAcceptorJob>(map, node.Id.ToString());
-            await scheduler.ScheduleJob(acceptorJob.Job, acceptorJob.Trigger);
+            node.JobId = job.Job.Key.ToString();
 
-            ToastService.ShowSuccess("Node subscription job created");
-        }
-        catch
-        {
-            ToastService.ShowError("Error while requesting to open the channel, please contact a superadmin for troubleshooting");
+            NodeRepository.Update(node);
```
RodriFS, line 455: you're missing checking the result:
```
var (result, _) = NodeRepository.Update(node);
if (!result)
{
    ToastService.ShowError($"Error while subscribing to job for node {node.Name}");
}

```

src/NodeGuard.csproj
```diff
@@ -59,7 +59,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <Folder Include="Migrations\" />
     <Folder Include="wwwroot\lib\" />
     <Folder Include="wwwroot\js\" />
   </ItemGroup>
```
RodriFS, line 64: you do need the migrations, sorry. File should be like it was before
```
  <ItemGroup>
    <Folder Include="Migrations\" />
    <Folder Include="wwwroot\lib\" />
    <Folder Include="wwwroot\lib\" />
    <Folder Include="wwwroot\js\" />
    <Folder Include="wwwroot\js\" />
  </ItemGroup>
```

RodriFS (approved):

PR #352, opened 2023-12-05
Null pointer exception when invalid metadata from btcpay

RodriFS:


markettes (approved):
LGTM. Although I would say it would be nice to have a message saying that it was invalid metadata.

src/Pages/Withdrawals.razor
```diff
@@ -586,23 +586,15 @@
...
+                    int atIndex = userName.IndexOf('@');
+                    return atIndex >= 0 ? userName.Substring(0, atIndex) : userName;
                 }
-            }
-            catch (JsonReaderException ex)
-            {
-                Console.WriteLine("Invalid JSON: " + ex.Message);
+                return userName;
             }
         }
```
AleksSoftware, line 600: I suggest improving the code to be more readable:

 return walletWithdrawalRequest.UserRequestor?.ToString() ?? string.Empty;

AleksSoftware:
LGTM🚀

src/Pages/Withdrawals.razor
Jossec101: Similar, defer to an issue.

PR #362, opened 2024-01-15
Withdrawalrequestspsbt can be stored more than once per signer

markettes:


src/Pages/ChannelRequests.razor
```diff
@@ -227,7 +227,13 @@
...
             <DataGridCommandColumn TItem="ChannelOperationRequest" Caption="Actions" Displayable="true">
                 <EditCommandTemplate>
-                    <Button Color="Color.Primary" hidden=@(!_isFinanceManager) Clicked="@(() => ShowModal(context.Item))">Approve</Button>
+                    @if (_isFinanceManager && context.Item.Wallet != null && context.Item.Wallet.Keys != null 
+                         && context.Item.Wallet.Keys.Count(x => x.UserId == LoggedUser?.Id) > context.Item.ChannelOperationRequestPsbts.Count(x => x.UserSignerId == LoggedUser?.Id) 
+                         && !context.Item.AreAllRequiredHumanSignaturesCollected)
+                    {
+                        <Button Color="Color.Primary" hidden=@(!_isFinanceManager) Clicked="@(() => ShowModal(context.Item))">Approve</Button>
+                    }
+
```
RodriFS, line 236: isn't this the job of `GetUnsignedPendingRequestsByUser`?

src/Pages/ChannelRequests.razor
```diff
@@ -795,7 +801,12 @@
 
         if (_selectedRequest == null || string.IsNullOrEmpty(_psbtSignRef?.SignedPSBT) || LoggedUser == null)
         {
-            ToastService.ShowError("Error: Not all fields were set");
+            ToastService.ShowError("Not all fields were set");
+        }
+        else if (_selectedRequest.ChannelOperationRequestPsbts.Any(x => x.UserSignerId == LoggedUser?.Id
+                                                                        && _psbtSignRef.SignedPSBT.Equals(x.PSBT)))
+        {
+            ToastService.ShowError("You already signed this PSBT");
```
RodriFS, line 809: i don't think this will work.
What if for a psbt with 3 needed signers
- user 1 signs, has two tabs open, signs on one of them
- user 2 signs
- user 1 clicks "approve" in second tab, the brought psbt will have the user 2 signature, but will be compared to the old one when user 1 signed.

Also, if you create a user with 2 keys, you can still sign with the same signature twice. I think you'll have to examine the psbt for the xpubs that signed

RodriFS (changes requested):

src/Pages/ChannelRequests.razor
markettes: From what I have tested, it works well, what you think is that you get the last signed PSBT when you click approve, but it prints the template:
```
try
{
    var templatePSBT = await BitcoinService.GenerateTemplatePSBT(walletWithdrawalRequest);

    //TODO Save template PSBT (?)
    _templatePsbtString = templatePSBT.ToBase64();
    if (_psbtSignRef != null) await _psbtSignRef.ShowModal();
}
```

src/Pages/ChannelRequests.razor
markettes: I was doing it the same way as in withdrawals. You are right!

markettes: I will do it and implement something similar for withdrawals for unifying both withdrawals and channel requests

markettes: Done 🏁

src/Pages/ChannelRequests.razor
RodriFS: I just tested, it doesn't let you sign more than the amount of signers, it gives you an error.
But you definitely can sign twice with the same key, and it will stay indefinitely in `Finalizing PSBT`

markettes: I removed the attribute giving some problems and checked for different users with the same keys. Done 🏁

RodriFS (approved):

PR #369, opened 2024-04-16
Add remote node pubkey field to LiquidityRules and protos

Jossec101:
This pull request adds a new field called `remote_node_pubkey` to the `LiquidityRule` class and the corresponding protobuf message. This field represents the public key of the node that is the counterparty of the channel.

This PR is needed to support automatic rebalancing without swaps.

This PR is needed for https://github.com/Elenpay/liquidator/pull/74

### DIFF
--- a/src/Proto/nodeguard.proto
+++ b/src/Proto/nodeguard.proto
@@ -80,6 +80,7 @@ message GetLiquidityRulesResponse {
 
 message LiquidityRule {
   uint64 channel_id = 1;
+  // Public key of the node that is the managed node of the channel
   string node_pubkey = 2;
   int32 swap_wallet_id = 3;
   optional int32 reverse_swap_wallet_id = 4;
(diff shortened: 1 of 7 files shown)


RodriFS (approved):

src/Proto/nodeguard.proto
```diff
@@ -88,6 +89,8 @@ message LiquidityRule {
   float rebalance_target = 7;
   optional string reverse_swap_address = 8;
   bool is_reverse_swap_wallet_rule = 9;
+  // Public key of the node that is the counterparty of the channel
+  string remote_node_pubkey = 10;
```
RodriFS, line 93: do you also send it if it's not managed by us?

Jossec101: Yes, they way this works is that liquidator will check both pubkeys, if they are managed, it will trigger a special rebalance. So yes, we need both pubkeys all the time.

PR #370, opened 2024-04-23
Cancel withdrawal request when no available utxos

RodriFS:
Related to https://github.com/Elenpay/btcpayserver/pull/182

### DIFF
--- a/src/Rpc/NodeGuardService.cs
+++ b/src/Rpc/NodeGuardService.cs
@@ -177,6 +177,17 @@ public class NodeGuardService : Nodeguard.NodeGuardService.NodeGuardServiceBase,
                 throw new RpcException(new Status(StatusCode.NotFound, "Wallet not found"));
             }
 
+            if (request.Amount <= 0)
+            {
+                throw new RpcException(new Status(StatusCode.InvalidArgument, "Amount must be greater than 0"));
+            }
+
+            if (request.Address == "")
+            {
+                throw new RpcException(new Status(StatusCode.InvalidArgument,
+                    "A destination address must be provided"));
+            }
+
             //Create withdrawal request
             var amount = new Money(request.Amount, MoneyUnit.Satoshi).ToUnit(MoneyUnit.BTC);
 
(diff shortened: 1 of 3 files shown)


src/Rpc/NodeGuardService.cs
```diff
@@ -239,9 +255,11 @@ IFMUTXORepository fmutxoRepository
                 _logger.LogError("Error saving withdrawal request for wallet with id {walletId}", request.WalletId);
                 throw new RpcException(new Status(StatusCode.Internal, "Error saving withdrawal request for wallet"));
             }
-            
+
             //Template PSBT generation with SIGHASH_ALL
-            var psbt = await _bitcoinService.GenerateTemplatePSBT(withdrawalRequest ?? throw new ArgumentException(nameof(withdrawalRequest)));
+            var psbt = await _bitcoinService.GenerateTemplatePSBT(withdrawalRequest ??
+                                                                  throw new ArgumentException(
+                                                                      nameof(withdrawalRequest)));
```
markettes, line 262: NIT: Strange line break

src/Rpc/NodeGuardService.cs
```diff
@@ -264,25 +282,41 @@ IFMUTXORepository fmutxoRepository
...
+            CancelWithdrawalRequest(withdrawalRequest);
             _logger.LogError("No available UTXOs for wallet with id {walletId}", request.WalletId);
-            throw new RpcException(new Status(StatusCode.Internal, "No available UTXOs for wallet"));
+            throw new RpcException(new Status(StatusCode.ResourceExhausted, "No available UTXOs for wallet"));
+        }
+        catch (NotEnoughBalanceInWalletException e)
+        {
+            _logger.LogError(e.Message);
+            throw new RpcException(new Status(StatusCode.ResourceExhausted, e.Message));
+        }
```
markettes, line 293: Doesn't it cancel when there are not enough funds?

markettes (approved):
LGTM just a question and a NIT

src/Rpc/NodeGuardService.cs
RodriFS: It's not needed because that error comes from the AddAsync. So there is no withdrawal request to cancel yet

src/Rpc/NodeGuardService.cs
RodriFS: it was like that before XD, i'll fix it

RodriFS: done!

src/Rpc/NodeGuardService.cs
markettes: Thanks!

PR #372, opened 2024-06-19
fix: utxo instead of fmutxos

markettes:


src/Migrations/20240619142749_remove-fm-references.cs
```diff
@@ -0,0 +1,188 @@
...
+
+namespace NodeGuard.Migrations
+{
+    /// <inheritdoc />
+    public partial class removefmreferences : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.DropTable(
```
Jossec101, line 15: Careful

Jossec101 (changes requested):
Dont drop the tables

PR #373, opened 2024-06-24
Utxo freezing data model

Jossec101:


Jossec101:
@markettes CLosing this, you can open it when its ready

PR #376, opened 2024-06-25
Data model for freezing UTXOs

markettes:
- Adds isFrozen and address to be used in the FMUTXOs
- Adds tags to FMUTXOs
- Adds a job to register all the new utxos in the system

```
- **feat: add address, isFrozen and tag**
- **feat: filter frozen utxos**
- **feat: utxo migration**
- **feat: add postgres for nbxplorer**
- **fix: add migration for optional address and tag**
- **feat: RegisterUTXOs job**
- **feat: add wallet to FMUTXO**
- **feat(FMUTXO.cs): add IsSpent property to determine if UTXO is spent based on channel and withdrawal statuses**
- **feat(FMUTXO.cs): add NBitcoin and NBXplorer dependencies for UTXO model enhancements**
- **feat: consolidate migrations**
- **fix: migration**
- **refactor(ChannelOperationRequestRepositoryTests): add additional null parameter to constructor calls to align with updated constructor signature**
- **Bump deps**
- **Fix problem with new change in nbxplorer when tracking btc addresses**
- **feat: rebuild of the migration**
- **EF Core already does execute the migration in a transaction scope, this was giving a race condition when generating the db (#375)**
- **feat: changes for deps update**
- **Fix AddRangeAsync not using a generic function type making EF Core fail to recognize the entities when saving a list**
- ** Fix  RegisterUTXOsJob to use WalletId instead of Wallet property**
- **fix: rollback changes**
- **feat(FMUTXO.cs): add Blazorise.Extensions for enhanced functionality refactor(FMUTXO.cs): improve IsSpent property logic to handle null collections feat(FUTXORepository.cs): include Tags in FMUTXOs query to provide more detailed data**
- **fix: remove error and set the creation time and the update time**
- **refactor(FUTXORepository.cs): optimize UTXO retrieval logic to enhance performance and accuracy feat(RegisterUTXOsJob.cs): log information after adding new UTXOs to enhance traceability fix(Constants.cs, launchSettings.json): align REGISTER_UTXOS_CRON schedule across configurations for consistency**

```


src/Data/Repositories/FUTXORepository.cs
```diff
@@ -50,6 +53,38 @@ public async Task<List<FMUTXO>> GetAll()
...
+                .ToListAsync();
+
+            List<FMUTXO> newUtxos = new List<FMUTXO>();
+            foreach (var utxo in allUtxos)
+            {
+                if (utxos.Contains(utxo))
+                {
+                    newUtxos.Add(utxo);
+                }
+            }
```
markettes, line 78: I've used this `foreach` because the linq call was not working for checking if the utxo checked was on the list. That `contains` call inside the linq call was not using our `equals`.

docker/docker-compose.yml
```diff
@@ -13,7 +13,7 @@ services:
     volumes:
       - nodeguard_postgres_data:/var/lib/postgresql/data
     ports:
-      - 25432:5432
+      - 5432:5432
```
RodriFS, line 16: i'd say let's keep nodeguard and our backend DBs separated

src/Data/Models/FMUTXO.cs
```diff
@@ -29,14 +34,31 @@ public class FMUTXO : Entity, IEquatable<FMUTXO>
...
+        
+        public int WalletId { get; set; }
+        public Wallet Wallet { get; set; }
 
         #endregion Relationships
+        
+        [NotMapped]
+        public bool IsSpent => (!ChannelOperationRequests.IsNullOrEmpty() || !WalletWithdrawalRequests.IsNullOrEmpty())
+                               && (ChannelOperationRequests.Any(x => x.Status == ChannelOperationRequestStatus.OnChainConfirmed) || 
+                                   WalletWithdrawalRequests.Any(x => x.Status == WalletWithdrawalRequestStatus.OnChainConfirmed));
```
RodriFS, line 61: What if it was spent outside of nodeguard?

src/Data/Models/FMUTXO.cs
```diff
@@ -55,6 +77,12 @@ public bool Equals(FMUTXO? other)
             return TxId == other.TxId && OutputIndex == other.OutputIndex && SatsAmount == other.SatsAmount;
         }
 
+        public bool Equals(UTXO? other)
+        {
+            if (ReferenceEquals(null, other)) return false;
+            return TxId == other.Outpoint.Hash.ToString() && OutputIndex == other.Outpoint.N && SatsAmount == ((Money)other.Value).Satoshi;
```
RodriFS, line 83: is `SatsAmount` needed here? I think if the hash and index are the same, then we're talking about the same utxo

src/Data/Repositories/FUTXORepository.cs
```diff
@@ -50,6 +53,38 @@ public async Task<List<FMUTXO>> GetAll()
...
 
+        public async Task<List<FMUTXO>> GetByWalletId(int walletId)
+        {
+            await using var applicationDbContext = await _dbContextFactory.CreateDbContextAsync();
+            
+            return await applicationDbContext.FMUTXOs.Where(x => x.Wallet.Id == walletId)
+                .Include(x => x.Tags).ToListAsync();
+        }
+
+        public async Task<List<FMUTXO>> GetFromUTXOs(List<FMUTXO> utxos)
```
RodriFS, line 64: this is a weird function, if you already have the list of utxos why do you need to get them from the db?

src/Data/Repositories/Interfaces/IRepository.cs
```diff
@@ -49,7 +49,7 @@ public interface IRepository<T>
         /// </summary>
         /// <param name="T"></param>
         /// <returns>A tuple (bool, string). The bool represents the call success and the string any possible message.</returns>
-        public Task<(bool, string?)> AddRangeAsync(List<T> T, ApplicationDbContext applicationDbContext);
+        public Task<(bool, string?)> AddRangeAsync<T>(List<T> entities, ApplicationDbContext applicationDbContext) where T : class;
```
RodriFS, line 52: nit. should change the generic name: 
```
Type parameter 'T' has the same name as the type parameter from outer type 'IRepository<T>'
```

src/Data/Repositories/Repository.cs
```diff
@@ -65,19 +65,20 @@ public async Task<(bool, string?)> AddAsync(T type, ApplicationDbContext applica
             return (rowsChanged, null);
         }
 
-        public async Task<(bool, string?)> AddRangeAsync(List<T> T, ApplicationDbContext applicationDbContext)
+        public async Task<(bool, string?)> AddRangeAsync<T>(List<T> entities, ApplicationDbContext applicationDbContext) where T : class
```
RodriFS, line 68: same

src/Jobs/RegisterUTXOsJob.cs
RodriFS, line 1: Can we instead fill the id of the wallet from the nbxplorer service when you call GetUTXOsAsync and GetUTXOsByLimitAsync?

docker/docker-compose.yml
markettes: Done 🏁

src/Data/Models/FMUTXO.cs
markettes: Good call; I don't think it's needed, tho. Will remove it.

src/Data/Models/FMUTXO.cs
markettes: You are right, It's probably not needed, but it's a good check anyway. Also, it is checked in the `public bool Equals(FMUTXO? other)` method.

src/Data/Repositories/FUTXORepository.cs
markettes: You are right, I will rethink this. Initially, this was planned to be a method to retrieve a set of utxos from the database directly instead of filtering at the code level.

src/Data/Repositories/Interfaces/IRepository.cs
markettes: Done 🏁

markettes: This produces a change to add to all the calls to this method. This change adds a lot of noise to the PR IMO. We can do it later.

src/Data/Repositories/Interfaces/IFMUTXORepository.cs
```diff
@@ -26,6 +27,10 @@ public interface IFMUTXORepository
     Task<FMUTXO?> GetById(int id);
 
     Task<List<FMUTXO>> GetAll();
+    
+    Task<List<FMUTXO>> GetByWalletId(int walletId);
+    
+    Task<FMUTXO> GetByOutpoint(string hash, uint outputIndex);
```
RodriFS, line 33:
```suggestion
    Task<FMUTXO?> GetByOutpoint(string hash, uint outputIndex);
```

src/Services/CoinSelectionService.cs
```diff
@@ -113,13 +115,52 @@ private IBitcoinRequestRepository GetRepository(BitcoinRequestType requestType)
...
+                _logger.LogError($"Could not save the utxos that were not selected for request:{bitcoinRequest.Id}");
+                utxosToAdd = new List<FMUTXO>();
+            }
+
+            foreach (var fmutxo in utxosToSave)
+            {
+                var finalUtxo = await _fmutxoRepository.GetByOutpoint(fmutxo.TxId, fmutxo.OutputIndex);
+                utxosToAdd.Add(finalUtxo);
+            }
+        }
```
RodriFS, line 152: i think you can achieve the same with a `Join`?

src/Services/CoinSelectionService.cs
```diff
@@ -113,13 +115,52 @@ private IBitcoinRequestRepository GetRepository(BitcoinRequestType requestType)
     public async Task LockUTXOs(List<UTXO> selectedUTXOs, IBitcoinRequest bitcoinRequest, BitcoinRequestType requestType)
     {
         // We "lock" the PSBT to the channel operation request by adding to its UTXOs collection for later checking
-        var utxos = selectedUTXOs.Select(x => _mapper.Map<UTXO, FMUTXO>(x)).ToList();
+        var selectedFMUTXOs = selectedUTXOs.Select(x => _mapper.Map<UTXO, FMUTXO>(x)).ToList();
+        var utxos = await _fmutxoRepository.GetByWalletId(bitcoinRequest.Wallet.Id);
 
-        var addUTXOsOperation = await GetRepository(requestType).AddUTXOs(bitcoinRequest, utxos);
+        var utxosToSave = new List<FMUTXO>();
+        var utxosToAdd = new List<FMUTXO>();
```
RodriFS, line 122: I think we could benefit from some comments here. And also add some TODOs if needed

RodriFS (approved):

src/Data/Repositories/ChannelOperationRequestRepository.cs
```diff
@@ -173,14 +176,7 @@ public async Task<(bool, string?)> AddUTXOs(IBitcoinRequest type, List<FMUTXO> u
...
                 if (request != null)
                 {
-                    if (!request.Utxos.Any())
-                    {
-                        request.Utxos = utxos;
-                    }
-                    else
-                    {
-                        request.Utxos.AddRange(utxos.Except(request.Utxos));
-                    }
```
Jossec101, line 183: Why has this been removed?

src/Data/Repositories/FUTXORepository.cs
```diff
@@ -17,9 +17,12 @@
  *
  */
 
-﻿using NodeGuard.Data.Models;
+using AutoMapper;
```
Jossec101, line 20:
```suggestion

```

src/Data/Repositories/FUTXORepository.cs
```diff
@@ -17,9 +17,12 @@
...
+using NodeGuard.Data.Models;
 using NodeGuard.Data.Repositories.Interfaces;
 using Microsoft.EntityFrameworkCore;
+using NBitcoin;
+using NBXplorer.Models;
```
Jossec101, line 25: Check unused usings

src/Data/Repositories/WalletWithdrawalRequestRepository.cs
```diff
@@ -209,14 +212,7 @@ public async Task<(bool, string?)> AddUTXOs(IBitcoinRequest type, List<FMUTXO> u
                     .SingleOrDefaultAsync(x => x.Id == type.Id);
                 if (request != null)
                 {
-                    if (!request.UTXOs.Any())
```
Jossec101, line 212: Same comment, why this removal?

src/Data/Models/FMUTXO.cs
```diff
@@ -29,15 +34,26 @@ public class FMUTXO : Entity, IEquatable<FMUTXO>
         public uint OutputIndex { get; set; }
 
         public long SatsAmount { get; set; }
-
+        
+        public string? Address { get; set; }
+        
+        public bool IsFrozen { get; set; }
```
Jossec101, line 40: Pls comment

src/Data/Models/FMUTXO.cs
```diff
@@ -29,15 +34,26 @@ public class FMUTXO : Entity, IEquatable<FMUTXO>
...
+        
+        public List<UTXOTag>? Tags { get; set; }
```
Jossec101, line 42: Move this to relationships #region

src/Data/Models/FMUTXO.cs
```diff
@@ -29,15 +34,26 @@ public class FMUTXO : Entity, IEquatable<FMUTXO>
         public uint OutputIndex { get; set; }
 
         public long SatsAmount { get; set; }
-
+        
+        public string? Address { get; set; }
```
Jossec101, line 38: bitcoin address? if so comment

src/Jobs/RegisterUTXOsJob.cs
```diff
@@ -0,0 +1,65 @@
...
+using NodeGuard.Data.Models;
+using NodeGuard.Data.Repositories.Interfaces;
+using NodeGuard.Helpers;
+using NodeGuard.Services;
+using Quartz;
+using ILogger = Microsoft.Build.Framework.ILogger;
+
+namespace NodeGuard.Jobs;
+
+public class RegisterUTXOsJob : IJob
```
Jossec101, line 12: Please doc this

src/Jobs/RegisterUTXOsJob.cs
```diff
@@ -0,0 +1,65 @@
...
+        _walletRepository = walletRepository;
+        _fmutxoRepository = fmutxoRepository;
+        _mapper = mapper;
+    }
+    public async Task Execute(IJobExecutionContext context)
+    {
+        var wallets = await _walletRepository.GetAll();
+        foreach (var wallet in wallets)
+        {
+            var utxos = await _nbXplorerService.GetUTXOsAsync(wallet.GetDerivationStrategy());
```
Jossec101, line 34: Potential null

Jossec101 (changes requested):

src/Services/CoinSelectionService.cs
Jossec101: +1 this is hard to read

src/Data/Repositories/Interfaces/IFMUTXORepository.cs
markettes: Done 🏁

docker/docker-compose.yml
```diff
@@ -28,8 +28,8 @@ services:
       NBXPLORER_TRIMEVENTS: 10000
       NBXPLORER_SIGNALFILESDIR: /datadir
       #Keeping dbtrie for dev until it is fully removed since we would need to modify nbxplorer docker image to wait for the db to be ready
-      NBXPLORER_DBTRIE: 1
-      # NBXPLORER_POSTGRES: Host=nodeguard_postgres;Port=5432;Database=nbxplorer;Username=rw_dev;Password=rw_dev
+      NBXPLORER_DBTRIE: 0
+      NBXPLORER_POSTGRES: Host=nbxplorer_postgres;Port=5432;Database=nbxplorer;Username=rw_dev;Password=rw_dev
```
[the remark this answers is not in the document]
RodriFS: we can create a `docker-compose.dev.yml` in the meantime with the dev config @markettes

src/Data/Repositories/Interfaces/IFMUTXORepository.cs
```diff
@@ -17,7 +17,8 @@
  *
  */
 
-﻿using NodeGuard.Data.Models;
+using NBXplorer.Models;
```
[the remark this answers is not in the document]
markettes: Done 🏁

src/Data/Repositories/FUTXORepository.cs
markettes: Done 🏁

src/Services/CoinSelectionService.cs
markettes: Done 🏁

docker/docker-compose.yml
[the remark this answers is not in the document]
markettes: I don't think its needed, we can just change the variable and uncomment the code if needed 😄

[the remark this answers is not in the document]
markettes: Done 🏁

src/Services/CoinSelectionService.cs
```diff
@@ -113,13 +115,58 @@ private IBitcoinRequestRepository GetRepository(BitcoinRequestType requestType)
...
     {
         // We "lock" the PSBT to the channel operation request by adding to its UTXOs collection for later checking
-        var utxos = selectedUTXOs.Select(x => _mapper.Map<UTXO, FMUTXO>(x)).ToList();
+        var selectedFMUTXOs = selectedUTXOs.Select(x => _mapper.Map<UTXO, FMUTXO>(x)).ToList();
+        var utxos = await _fmutxoRepository.GetByWalletId(bitcoinRequest.Wallet.Id);
 
-        var addUTXOsOperation = await GetRepository(requestType).AddUTXOs(bitcoinRequest, utxos);
+        var utxosToSave = new List<FMUTXO>();
+        var utxosToAddToTransaction = new List<FMUTXO>();
+        // We check the selected UTXOs against the ones we have in the database to see if they are already there
```
RodriFS, line 123: nit:
```suggestion
        // Since the UTXOs to lock (selectedUTXOs) come directly from nbxplorer, we check if we already registered them in our database with the RegisterUTXOsJob
```

src/Services/CoinSelectionService.cs
```diff
@@ -113,13 +115,58 @@ private IBitcoinRequestRepository GetRepository(BitcoinRequestType requestType)
...
+            (bool success, string? reason) = await _fmutxoRepository.AddRangeAsync(utxosToSave);
+            if (!success)
+            {
+                _logger.LogError($"Could not save the utxos that were not selected for request:{bitcoinRequest.Id}");
+                utxosToAddToTransaction = new List<FMUTXO>();
+            }
+
+            foreach (var fmutxo in utxosToSave)
+            {
+                // We get the final UTXO from the DB to add to the transaction
```
RodriFS, line 153: nit: this is the reason, right?
```suggestion
                // We refetch the UTXOs from the DB so we can have them with the id
```

src/Services/CoinSelectionService.cs
```diff
@@ -113,13 +115,58 @@ private IBitcoinRequestRepository GetRepository(BitcoinRequestType requestType)
...
+                var finalUtxo = await _fmutxoRepository.GetByOutpoint(fmutxo.TxId, fmutxo.OutputIndex);
+                utxosToAddToTransaction.Add(finalUtxo);
+            }
+        }
+
+        // If we couldn't save the UTXOs we fail to construct the transaction
+        if (utxosToAddToTransaction.Count < selectedUTXOs.Count)
+        {
+            utxosToAddToTransaction = new List<FMUTXO>();
+        }
```
RodriFS, line 163: I'm still not sure if we should throw or not. @Jossec101?

src/Jobs/RegisterUTXOsJob.cs
markettes: Fixed 🏁

src/Data/Models/FMUTXO.cs
markettes: Done 🏁

src/Data/Models/FMUTXO.cs
markettes: Done 🏁

src/Data/Models/FMUTXO.cs
markettes: Done 🏁

src/Data/Repositories/ChannelOperationRequestRepository.cs
markettes: This is unnecessary since we are taking the FMUTXOs from already saved ones. Previously, we wanted this, so we don't save many times the same UTXO, as far as I know.

RodriFS:
@markettes @Jossec101 at some point we should remove the utxos from the table once they are spent. Otherwise the table will grow forever. should we make a ticket?

markettes:
> @markettes @Jossec101 at some point we should remove the utxos from the table once they are spent. Otherwise the table will grow forever. should we make a ticket?

That's what we had already, but I agree, we should make a ticket for that

RodriFS (dismissed):

RodriFS (dismissed):

RodriFS (dismissed):

src/Helpers/Constants.cs
```diff
@@ -85,7 +85,9 @@ public class Constants
     /// Max ratio of the tx total input sum that could be used as fee
     /// </summary>
     public static decimal MAX_TX_FEE_RATIO =0.5m;
-
+    
+    public static string IsFrozenTag = "frozen";
```
Jossec101, line 89:
```suggestion
    public static string IS_FROZEN_TAG = "frozen";
```

Jossec101 (dismissed):

Jossec101 (approved):

RodriFS (approved):

PR #377, opened 2024-06-26
Add address to GetAvailableUtxos in NodeguardService

RodriFS:


markettes (approved):
LGTM

src/Rpc/NodeGuardService.cs
```diff
@@ -868,6 +870,42 @@ private bool ValidateTargetBalance(LiquidityRule rule)
...
 
+    public override async Task<GetAllUtxosResponse> GetAllUtxos(GetAllUtxosRequest request, ServerCallContext context)
+    {
+        List<FMUTXO> fmutxos;
+        // Check if the user sent a wallet id, else we get all utxos
+        if (request.WalletId == 0)
+        {
+            fmutxos = await _fmutxoRepository.GetAll();
+        }
+        else
```
daliclovr, line 881: why are we filtering by wallet? 🤔

src/Rpc/NodeGuardService.cs
```diff
@@ -868,6 +870,42 @@ private bool ValidateTargetBalance(LiquidityRule rule)
...
+        {
+            fmutxos = await _fmutxoRepository.GetByWalletId(request.WalletId);
+        }
+
+        // Check if user wants the list to contain the locked utxos
+        if (!request.IncludeLocked)
+        {
+            var lockedUtxos = await _fmutxoRepository.GetLockedUTXOs();
+            Console.WriteLine(lockedUtxos.Count);
+            Console.WriteLine(fmutxos.Count);
```
daliclovr, line 891: Writeline 🐔

src/Rpc/NodeGuardService.cs
```diff
@@ -868,6 +870,42 @@ private bool ValidateTargetBalance(LiquidityRule rule)
...
+            fmutxos = fmutxos.Except(lockedUtxos).ToList();
+        }
+
+        var utxos = fmutxos.Select(utxo => new Utxo()
+        {
+            Amount = new Money(utxo.SatsAmount),
+            // We follow the same convention as GetAvailableUtxos for displaying outpoints
+            Outpoint = $"{utxo.TxId}-{utxo.OutputIndex}",
+            Address = utxo.Address?.ToString()
```
daliclovr, line 900: can you explain to me why an address could be null?

daliclovr:
@RodriFS can you explain to me what is considered an `AvailableUTXO`? I imagine that is something like "your utxo is not spent but is being used (i,e lightning channel)". Are unavailablesUTXO marked as locked? how are they marked?

daliclovr (commented):
@RodriFS can you explain to me what is considered an AvailableUTXO? I imagine that is something like "your utxo is not spent but is being used (i,e lightning channel)". Are unavailablesUTXO marked as locked? how are they marked?

src/Rpc/NodeGuardService.cs
RodriFS: the default case is that it brings all utxos no matter if the wallet is compromised. So i'm giving the option to the user to first fetch the available wallets and then get the utxos for those ones. I could also add a flag for bringing unarchived/uncompromised wallets

src/Rpc/NodeGuardService.cs
RodriFS: removed!

src/Rpc/NodeGuardService.cs
RodriFS: it shouldn't. I think it is nullable because it's a new row and it is not filled by default. @markettes where do we go back and fill this for utxos already on the table?

src/Rpc/NodeGuardService.cs
RodriFS: Actually the functions `GetAvailableUtxos` and `GetAllUtxos` should be one function, but since `GetAvailableUtxos` is used by our backend to do settlements, I didn't want to modify it just yet. There is a difference between those two:

The `GetAvailableUtxos` function gets the utxos from the nbxplorer, then we remove the locked/frozen utxos. This means that that if there are utxos that haven't been tagged just yet, the caller to the function is going to get those utxos anyway.

The `GetAllUtxos` only returns utxos that have been already registered by us with the `RegusterUTXOsJob`.

- We have two things to do in the future: Add a flag to `GetAvailableUtxos` so we can choose to return only tagged utxos
- Merge both into the `GetAllUtxos` function and deprecate `GetAvailableUtxos`

src/Rpc/NodeGuardService.cs
markettes: We don't really need to fill these old UTXOs, we could just remove the data from this table and it will be generated.

src/Rpc/NodeGuardService.cs
```diff
@@ -868,6 +870,28 @@ private bool ValidateTargetBalance(LiquidityRule rule)
...
+        var wallets = await _walletRepository.GetAvailableWallets();
+        List<Utxo> utxos = new();
+        foreach (var wallet in wallets)
+        {
+            var walletUtxos = await _fmutxoRepository.GetByWalletId(wallet.Id);
+            utxos.AddRange(walletUtxos.Select(utxo => new Utxo()
+            {
+                Amount = new Money(utxo.SatsAmount),
+                // We follow the same convention as GetAvailableUtxos for displaying outpoints
+                Outpoint = $"{utxo.TxId}-{utxo.OutputIndex}",
```
daliclovr, line 884: you already have this in the model!
```
        public override string ToString()
        {
            return $"{TxId}:{OutputIndex}";
        }
```

src/Rpc/NodeGuardService.cs
```diff
@@ -901,15 +925,20 @@ public override async Task<GetAvailableUtxosResponse> GetAvailableUtxos(GetAvail
             lockedUtxos.Select(utxo => $"{utxo.TxId}-{utxo.OutputIndex}").ToList()
             );
 
+        // TODO: check if they are not registered
+        // go to database, get all utxos, compare and remove the ones that are not registered/tagged
```
daliclovr, line 929: ure not doing this at the end?

daliclovr (changes requested):

src/Rpc/NodeGuardService.cs
daliclovr: u're doing this in another task, right?

src/Rpc/NodeGuardService.cs
daliclovr: cool

daliclovr:
> > @RodriFS can you explain to me what is considered an AvailableUTXO? I imagine that is something like "your utxo is not spent but is being used (i,e lightning channel)". Are unavailablesUTXO marked as locked? how are they marked?
> 
> That's exactly right. Avaliable utxos are locked in a request or frozen. There is a different on how we manage locked and frozen utxos:
> 
> * the locked utxos are not marked, each time you request the list of available utxos, we go and check if there is a request that is taking hold of that utxos and we don't add those to the list
> * the frozen utxos are marked with a flag in the database
> 
> You can see the implementation here:
> 
> https://github.com/Elenpay/NodeGuard/blob/69c1acfceb39f7d5e619f7f9bee6f889b69c1591/src/Data/Repositories/FUTXORepository.cs#L108

@RodriFS I feel that what you clarify to me today, the difference between `locked` and `frozen` should be reflected somewhere. Both status means the same in the sense of `utxos` are not usable if locked or frozen, although `frozen` is a "permanent" status and locked is a computed status based on others

src/Rpc/NodeGuardService.cs
RodriFS: yes

src/Rpc/NodeGuardService.cs
RodriFS: we follow the convention of the nbxplorer, so we have to do it with a dash instead of the colon

src/Rpc/NodeGuardService.cs
RodriFS: yes, @markettes did it in his task

RodriFS:
> > > @RodriFS can you explain to me what is considered an AvailableUTXO? I imagine that is something like "your utxo is not spent but is being used (i,e lightning channel)". Are unavailablesUTXO marked as locked? how are they marked?
> > 
> > 
> > That's exactly right. Avaliable utxos are locked in a request or frozen. There is a different on how we manage locked and frozen utxos:
> > 
> > * the locked utxos are not marked, each time you request the list of available utxos, we go and check if there is a request that is taking hold of that utxos and we don't add those to the list
> > * the frozen utxos are marked with a flag in the database
> > 
> > You can see the implementation here:
> > https://github.com/Elenpay/NodeGuard/blob/69c1acfceb39f7d5e619f7f9bee6f889b69c1591/src/Data/Repositories/FUTXORepository.cs#L108
> 
> @RodriFS I feel that what you clarify to me today, the difference between `locked` and `frozen` should be reflected somewhere. Both status means the same in the sense of `utxos` are not usable if locked or frozen, although `frozen` is a "permanent" status and locked is a computed status based on others. Is quite confusing if this information is not added anywhere...

@daliclovr please review the PR again. We removed the IsFrozen flag on the utxos, and now we just use the tags to know the status of a utxo. So we don't have to register utxos in the table at all, we just tag utxos comming from the nbxplorer

src/Rpc/NodeGuardService.cs
markettes: Now, this `TODO` is not needed

markettes (dismissed):

src/Rpc/NodeGuardService.cs
RodriFS: removed

src/Proto/nodeguard.proto
```diff
@@ -307,6 +312,11 @@ enum COIN_SELECTION_STRATEGY
     UP_TO_AMOUNT = 3;
 }
 
+message GetUtxosRequest {
+  // Wallet ID as stored in the NG's database
+  optional int32 wallet_id = 1;
```
Jossec101, line 317: This field is not used in the RPC API

src/Rpc/NodeGuardService.cs
```diff
@@ -871,7 +886,43 @@ private bool ValidateTargetBalance(LiquidityRule rule)
         return true;
     }
 
-    public override async Task<GetAvailableUtxosResponse> GetAvailableUtxos(GetAvailableUtxosRequest request, ServerCallContext context)
+    public override async Task<GetUtxosResponse> GetUtxos(GetUtxosRequest request, ServerCallContext context)
```
Jossec101, line 889: Request is not used, so this is in fact, `GetAllUTXOs`

src/Proto/nodeguard.proto
```diff
@@ -57,7 +57,12 @@ service NodeGuardService {
   /*
   Gets a list of available UTXOs for a wallet
    */
-  rpc GetAvailableUtxos(GetAvailableUtxosRequest) returns (GetAvailableUtxosResponse);
+  rpc GetAvailableUtxos(GetAvailableUtxosRequest) returns (GetUtxosResponse);
+
+  /*
+  Gets a list of UTXOs
```
Jossec101, line 63: If this is all the UTXOs in the treasury, please doc it

src/Proto/nodeguard.proto
RodriFS: removed

src/Proto/nodeguard.proto
RodriFS: documented better

src/Rpc/NodeGuardService.cs
RodriFS: created issue https://github.com/Elenpay/NodeGuard/issues/382

Jossec101 (dismissed):

src/Rpc/NodeGuardService.cs
RodriFS: https://github.com/Elenpay/NodeGuard/issues/382

daliclovr:
> > > > @RodriFS can you explain to me what is considered an AvailableUTXO? I imagine that is something like "your utxo is not spent but is being used (i,e lightning channel)". Are unavailablesUTXO marked as locked? how are they marked?
> > > 
> > > 
> > > That's exactly right. Avaliable utxos are locked in a request or frozen. There is a different on how we manage locked and frozen utxos:
> > > 
> > > * the locked utxos are not marked, each time you request the list of available utxos, we go and check if there is a request that is taking hold of that utxos and we don't add those to the list
> > > * the frozen utxos are marked with a flag in the database
> > > 
> > > You can see the implementation here:
> > > https://github.com/Elenpay/NodeGuard/blob/69c1acfceb39f7d5e619f7f9bee6f889b69c1591/src/Data/Repositories/FUTXORepository.cs#L108
> > 
> > 
> > @RodriFS I feel that what you clarify to me today, the difference between `locked` and `frozen` should be reflected somewhere. Both status means the same in the sense of `utxos` are not usable if locked or frozen, although `frozen` is a "permanent" status and locked is a computed status based on others. Is quite confusing if this information is not added anywhere...
> 
> @daliclovr please review the PR again. We removed the IsFrozen flag on the utxos, and now we just use the tags to know the status of a utxo. So we don't have to register utxos in the table at all, we just tag utxos comming from the nbxplorer

@RodriFS how can you tag the utxos in advance without knowing the score? What is the status to consider them `locked` now?

daliclovr (approved):

PR #381, opened 2024-06-28
RPC call for adding UTXOTags

markettes:


src/Rpc/NodeGuardService.cs
```diff
@@ -994,7 +996,45 @@ public override async Task<GetChannelResponse> GetChannel(GetChannelRequest requ
...
+            if (utxo.Length != 2)
+            {
+                throw new RpcException(new Status(StatusCode.InvalidArgument, $"Invalid outpoint {tag.UtxoOutpoint}"));
+            }
+
+            if (!uint.TryParse(utxo[1], out _))
+            {
+                throw new RpcException(new Status(StatusCode.InvalidArgument, $"Invalid output index {tag.UtxoOutpoint}"));
+            }
+        }
```
RodriFS, line 1019: you can validate the outpoint directly with `OutPoint.TryParse()`

RodriFS (dismissed):

src/Proto/nodeguard.proto
```diff
@@ -68,6 +68,11 @@ service NodeGuardService {
   Gets a channel by id
    */
   rpc GetChannel(GetChannelRequest) returns (GetChannelResponse);
+  
+  /*
+  Adds tags
```
Jossec101, line 73:
```suggestion
  Adds tags to UTXOs in the system as
```

src/Rpc/NodeGuardService.cs
```diff
@@ -994,7 +996,45 @@ public override async Task<GetChannelResponse> GetChannel(GetChannelRequest requ
 
         return result;
     }
-    
+
+    public async Task<AddTagsResponse> AddTags(AddTagsRequest request, ServerCallContext context)
```
Jossec101, line 1000:
```suggestion
    public override async Task<AddTagsResponse> AddTags(AddTagsRequest request, ServerCallContext context)
```

src/Rpc/NodeGuardService.cs
Jossec101: @RodriFS can you implement that?

Jossec101 (approved):

PR #384, opened 2024-07-03
Fixed Cryptographic exception

RodriFS:
It looks like the culprit for the exception was the ProtecetedLocalStorage. Since the keys have to be protected with
the data protection key, when the key id was not found in the database, the local storage crashed. Now if the key is
not found in the DB, we set the storage with the new data protection key


Jossec101 (approved):
Nice finding

PR #385, opened 2024-07-03
added elenpay for dev as api token

daliclovr:
This is to add a fixed dummy token for development called ElenPay

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -22,7 +22,7 @@ services:
     platform: linux/amd64
     hostname: nbxplorer
     ports:
-      - "32838:32838"
+    - "32838:32838"
     environment:
       NBXPLORER_NETWORK: regtest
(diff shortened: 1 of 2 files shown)


docker/docker-compose.yml
```diff
@@ -22,24 +22,24 @@ services:
     platform: linux/amd64
     hostname: nbxplorer
     ports:
-    - "32838:32838"
```
Jossec101, line 25: Pls dont modify the docker-compose, Im seeing all PRs changing it, let's sit down if needed but I would remove it from the PR

Jossec101 (changes requested):

docker/docker-compose.yml
daliclovr: restored

RodriFS (approved):

Jossec101 (approved):

PR #389, opened 2024-07-25
added reference_id to withdrawal request creation

daliclovr:
This adds `reference_id` field to RPC method `RequestWithdrawalRequest`. This is needed for settlements as we need to track our inner reference ids for the system (external reference id).

### DIFF
--- a/src/Proto/nodeguard.proto
+++ b/src/Proto/nodeguard.proto
@@ -132,7 +132,7 @@ message RequestWithdrawalRequest {
   FEES_TYPE mempool_fee_rate = 8;
   // Fee rate in sat/vbyte
   optional int32 custom_fee_rate = 9;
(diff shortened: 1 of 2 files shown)


src/Proto/nodeguard.proto
```diff
@@ -132,6 +132,8 @@ message RequestWithdrawalRequest {
   FEES_TYPE mempool_fee_rate = 8;
   // Fee rate in sat/vbyte
   optional int32 custom_fee_rate = 9;
+  // Exteneral reference id for the withdrawal request
```
qustavo, line 135: typo

qustavo (approved):

src/Proto/nodeguard.proto
daliclovr: solved

qustavo (approved):

PR #403, opened 2025-03-13
Add number of confirmations to withdrawal request status

RodriFS:


src/Rpc/NodeGuardService.cs
```diff
@@ -1033,33 +1033,41 @@ private WITHDRAWAL_REQUEST_STATUS GetStatus(WalletWithdrawalRequestStatus status
         };
     }
 
-    private GetWithdrawalsRequestStatusResponse GetWithdrawalsRequestStatusResponse(List<WalletWithdrawalRequest>? withdrawalRequests)
+    private async Task<GetWithdrawalsRequestStatusResponse> GetWithdrawalsRequestStatusResponse(List<WalletWithdrawalRequest> withdrawalRequests)
```
qustavo, line 1036: dumb question: why does this need to change to async task?

RodriFS: `Task` is always needed as a return parameter to await for an async call. You can't for example do `await GetSomething()` if the signature is `void` instead of `Task`

qustavo: weren't we already performing an async task when querying the db?

RodriFS: not inside the function `GetWithdrawalsRequestStatusResponse`

qustavo (approved):

PR #426, opened 2025-09-04
[GEN-1670] Configuration for loop + tilt

RodriFS:
This is a refactor of the nodes configuration so it works without polar
When the nodes are up, they create their own wallet and then they share this credentials with loop and loopserver.
When nodeguard goes up with `just run` or `just watch` it fetches the macaroons from the nodes and set them up in an env file that later reads.


docker/.justfile
```diff
@@ -0,0 +1,130 @@
...
+                ;;
+            "custom")
+                cmd=$(gum input --placeholder "Enter custom bitcoin-cli command")
+                just bitcoin-cli $cmd
+                ;;
+        esac
+    
+    elif [ "$program" = "lnd" ]; then
+        # LND commands
+        command=$(gum choose "getinfo" "addinvoice" "payinvoice" "listchannels" "openchannel" "closechannel" "walletbalance" "channelbalance" "newaddress" "custom")
```
imclvr, line 49: nitpick: I've seen `gum` being run through `go tool` in other places of the source code. Would be better to have an unique source for this tool and use it, IMHO.

Another alternative could be to add `gum` in `just install`, but I think this would make gum to be fetched from two sides, and better to have a single one.

README.md
```diff
@@ -72,27 +72,38 @@ dotnet tool install -g dotnet-ef
 
 ## Developing
 
-### Visual Studio
-Launch the FundsManager Docker VS task
-Launch The FundsManager VS task
-
-### Rider/IntelliJ
```
imclvr, line 79: Overall question that does not affect this PR: 

Have we considered to use a generic editor config, instead of being bound to VSCode? https://editorconfig.org/

docker/loop/docker-compose.loop.yml
Jossec101, line 1: NIT: if its under loop dir, why not name it `docker-compose.yml`

docker/mempool/docker-compose.mempool.yml
Jossec101, line 1: NIT: if its under its own dir, why not name it docker-compose.yml

Jossec101 (approved):

src/Data/Models/Node.cs
```diff
@@ -62,6 +62,10 @@ public class Node : Entity
         [NotMapped]
         public bool IsManaged => Endpoint != null;
 
+        public string? LoopEndpoint { get; set; }
+
+        public string? LoopMacaroon { get; set; }
+
```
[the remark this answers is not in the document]
RodriFS: we don't need them. neither the LND ones. The grpc server was always configured to trust anything

docker/loop/docker-compose.loop.yml
RodriFS: renamed

docker/mempool/docker-compose.mempool.yml
RodriFS: renamed

src/Data/Models/Node.cs
```diff
@@ -62,6 +62,10 @@ public class Node : Entity
         [NotMapped]
         public bool IsManaged => Endpoint != null;
 
+        public string? LoopEndpoint { get; set; }
```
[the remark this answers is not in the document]
RodriFS: renamed

PR #430, opened 2025-09-17
editorconfig

RodriFS:
.editorconfig taken from https://github.com/AzureAD/microsoft-identity-web/
I had to change some rules from error to warn, otherwise it wouldn't compile

Also, `dotnet format` doesn't format `.razor` files, so that is out of this PR. But it should work if you save manually


Jossec101 (commented):
@RodriFS what's the status of this?

PR #433, opened 2025-09-24
Add loopd tls cert

RodriFS:


docker/extract-macaroons.sh
```diff
@@ -215,6 +215,43 @@ fi
...
+    fi
+else
+    echo "# BOB_LOOPD_TLS_CERT=\"<container_not_running>\"" >> "${OUTPUT_FILE}"
+    echo -e "${YELLOW}⚠ Bob Loopd container not running${NC}"
+fi
+
+echo "" >> "${OUTPUT_FILE}"
+
+# Carol Loopd TLS
+echo "# Carol Loopd TLS Certificate" >> "${OUTPUT_FILE}"
```
imclvr, line 239: Suggestion: It feels this is more or less the same as with loopd service for Bob. Perhaps we should have a single function that can be called both for Bob and Carol, wdyt?

src/Migrations/20250924095441_AddLoopdCert.cs
```diff
@@ -0,0 +1,28 @@
...
+namespace NodeGuard.Migrations
+{
+    /// <inheritdoc />
+    public partial class AddLoopdCert : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<string>(
+                name: "LoopdCert",
```
imclvr, line 14: Concern/question: Perhaps this column should be encrypted at DB level, so that it's resilient to any DB breach.

Don't know how to do that with this ORM tbh, would require investigation from my side. What I know, Postgres has the `pgcrypto` module that could be handy:

https://www.postgresql.org/docs/current/pgcrypto.html

Jossec101 (approved):

src/Migrations/20250924095441_AddLoopdCert.cs
imclvr: No idea yet on how DB is configured tbh, but full encryption from configuration makes sense indeed.

PR #500, opened 2026-04-28
[GEN-524] Channel balances disrupted by disabled nodes

imclvr:


src/Services/LightningService.cs
imclvr, line 1424: This call is internally handling errors, catching an exception, logging it and then returning error. 
So I believe we are safe from what the [ticket](https://www.notion.so/Channel-balances-disrupted-by-disabled-nodes-32442ded9a6c82b9ad1c013823f93d6c?source=copy_link) says

```
Missing error handling in the UI: When lightning node returns an error or unexpected response, the lack of a try/catch block causes the entire balance rendering to break, preventing the user from seeing any channel balance information.
```

manumonti (approved):
LGTM! 👍

Jossec101 (approved):

PR #504, opened 2026-05-06
Add methods to manage channel fee policies and retrieve channels by outpoint

markettes:
Introduce methods for setting channel fee policies and retrieving channels by their outpoint. Implement unit tests to ensure functionality and correctness of these new features.

### DIFF
--- a/src/Proto/nodeguard.proto
+++ b/src/Proto/nodeguard.proto
@@ -89,6 +89,11 @@ service NodeGuardService {
   Adds tags to UTXOs under the treasury
    */
   rpc AddTags(AddTagsRequest) returns (AddTagsResponse);
+  
+  /*
+  Sets the fee policy for a channel
+   */
+  rpc SetChannelFeePolicy(SetChannelFeePolicyRequest) returns (SetChannelFeePolicyResponse);
 }
 
 message GetLiquidityRulesRequest {
@@ -443,4 +448,21 @@ message AddTagsRequest {
 }
 
 message AddTagsResponse {
+}
+
+message SetChannelFeePolicyRequest {
+  string chan_point = 1;
+  int64 base_fee_msat = 2;
+  uint32 fee_rate_ppm = 3;
+  uint32 time_lock_delta = 4;
+  optional InboundFeePolicy inbound_fee_policy = 5;
+  string node_pubkey = 6;
+}
+
+message InboundFeePolicy {
+  int32 base_fee_msat = 1;
+  int32 fee_rate_ppm = 2;
+}
+
+message SetChannelFeePolicyResponse {
 }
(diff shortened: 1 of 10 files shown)


src/Proto/nodeguard.proto
```diff
@@ -444,3 +449,20 @@ message AddTagsRequest {
 
 message AddTagsResponse {
 }
+
+message SetChannelFeePolicyRequest {
```
Jossec101, line 468: pls doc each field?

src/Services/LightningService.cs
```diff
@@ -1528,5 +1542,91 @@ public async Task<Dictionary<ulong, ChannelState>> GetChannelsState()
...
+                throw new ArgumentException("Node pubkey is required.", nameof(nodePubKey));
+            }
+
+            // Validate inbound fee policy parameters
+            if (inboundBaseFeeMsat.HasValue != inboundFeeRatePpm.HasValue)
+            {
+                throw new ArgumentException("Both inboundBaseFeeMsat and inboundFeeRatePpm must be provided together for inbound fee policy.");
+            }
+
+            var channel = await _channelRepository.GetByOutpoint(outPoint);
```
Jossec101, line 1848: Another validation is that `inboundFeeRatePpm` needs to be <= normal fee ppm rate. Im not sure if LND validates this but we can do also for defensive programming

src/Proto/nodeguard.proto
markettes: done

src/Services/LightningService.cs
markettes: done

src/Services/LightningService.cs
```diff
@@ -1801,5 +1815,109 @@ public async Task<Payment> SendPaymentV2Async(Node node, string paymentRequest,
...
+
+            // Validate inbound fee policy parameters
+            if (inboundBaseFeeMsat.HasValue != inboundFeeRatePpm.HasValue)
+            {
+                throw new ArgumentException("Both inboundBaseFeeMsat and inboundFeeRatePpm must be provided together for inbound fee policy.");
+            }
+
+            if (inboundBaseFeeMsat.HasValue && inboundFeeRatePpm.HasValue)
+            {
+                if (inboundBaseFeeMsat.Value > baseFeeMsat)
```
Jossec101: Sorry I made a mistake, it has to be lower in absolute values, also are we handling that these values should be always negative?

markettes: the part that inbound has to be ABS lower than outbound/general is not really accurate. It can be abs(normal)<abs(inbound) since inbound has nothing to do with outbound. Being inbound the channel x and y the outbound channel, the discound in inbound fees of channel x have to be lower than the outbound of channel y. Thats the correct constraint, hard to limit tbh

Jossec101: 100% true, solving.

Jossec101 (approved):

PR #506, opened 2026-05-07
Add Rebalance data model

Jossec101:
Stacked PRs:
 * #510
 * #509
 * #508
 * #507
 * __->__#506

--- --- ---

### Add Rebalance data model

- Created a new migration for the Rebalances table with relevant fields
- Updated the ApplicationDbContextModelSnapshot to include the Rebalance entity configuration,

### DIFF
--- a/src/Data/Models/Rebalance.cs
+++ b/src/Data/Models/Rebalance.cs
@@ -1,0 +1,135 @@
+/*
+ * NodeGuard
+ * Copyright (C) 2023  Elenpay
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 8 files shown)


src/Data/Models/Rebalance.cs
```diff
@@ -0,0 +1,135 @@
...
+
+    /// <summary>
+    /// Maximum fee as percentage of <see cref="SatsAmount"/> (0.05 = 0.05%).
+    /// Escalates on retry per <see cref="RetryMaxFeePct"/> or Constants.
+    /// </summary>
+    public double MaxFeePct { get; set; }
+    /// <summary>
+    /// Maximum fee percent done in retries, falling to a user-supplied value or defaulting to Constants a retry is needed.
+    /// </summary>
+    public double? RetryMaxFeePct { get; set; }
```
markettes, line 79: Why would we want to set a different max fee in case we retry?

Jossec101: Its purely optional, you can set the same but it avoids having to send another request if you want to maximize success over anything

markettes (approved):

PR #508, opened 2026-05-07
Rebalancing Job + Required services

Jossec101:
Stacked PRs:
 * #510
 * #509
 * __->__#508

--- --- ---

### Rebalancing Job + Required services

The idea is to queryroute + probe before using real SendPayment, during probing the amount is reduced by a backoff factor if there's no real liquidity, SendPayment is used to use MPP easier and has built-in retries.

### DIFF
--- a/src/Services/LightningService.cs
+++ b/src/Services/LightningService.cs
@@ -142,6 +142,49 @@ namespace NodeGuard.Services
         /// <param name="timeout">Optional timeout in seconds</param>
         /// <returns></returns>
         public Task<RouteFeeResponse?> EstimateRouteFee(string destPubkey, long amountSat, string? paymentRequest = null, uint timeout = 30);
+
+        /// <summary>
+        /// Adds a self-payable invoice on the given node. Used for circular rebalancing where
+        /// the source node pays itself
+        /// </summary>
+        Task<AddInvoiceResponse?> AddInvoiceAsync(Node node, long amountSats, string memo, long expirySeconds);
+
+        /// <summary>
+        /// Probes a candidate route at the requested amount, reducing the amount until either a route
+        /// works (signaled by IncorrectOrUnknownPaymentDetails at the last hop) or the
+        /// minimum probe amount is reached. Returns the probed amount and the validated
+        /// route, or NoRoute if no route was found.
+        /// </summary>
+        Task<ProbeResult> ProbeRouteAsync(Node node, long amountSats, long feeLimitMsat,
+            ulong? outgoingChanId, string? lastHopPubkeyHex, double probeBackoffRatio, CancellationToken ct);
+
+        /// <summary>
+        /// Sends a self-payment via Routerrpc.SendPaymentV2 with the given constraints and
+        /// drains the streaming response to the terminal Payment update. LND drives MPP and
+        /// internal route iteration / retries; we provide the cost cap (feeLimitMsat) and
+        /// the optional last-hop peer pin (lastHopPubkeyHex; null lets LND pick any inbound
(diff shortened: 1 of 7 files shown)


src/Services/LightningService.cs
```diff
@@ -1528,5 +1571,186 @@ public async Task<Dictionary<ulong, ChannelState>> GetChannelsState()
...
+                Value = amountSats,
+                Expiry = expirySeconds,
+                Private = false,
+            });
+        }
+
+        public async Task<ProbeResult> ProbeRouteAsync(Node node, long amountSats, long feeLimitMsat,
+            ulong? outgoingChanId, string? lastHopPubkeyHex, double probeBackoffRatio, CancellationToken ct)
+        {
+            // Ratio must be in the open interval (0, 1). 0 zeroes the next try; 1 never shrinks.
```
markettes: (0,1] might be useful to retry with the same amount

src/Services/LightningService.cs
```diff
@@ -1528,5 +1571,186 @@ public async Task<Dictionary<ulong, ChannelState>> GetChannelsState()
...
+                    {
+                        _logger.LogWarning(e, "Probe SendToRouteV2 failed for node {NodeId} amount {Amount}", node.Id, amount);
+                        continue;
+                    }
+
+                    if (attempt?.Failure?.Code == Failure.Types.FailureCode.IncorrectOrUnknownPaymentDetails)
+                    {
+                        // The route worked all the way to the destination; the random hash
+                        // failed at the final hop, which is the probe success signal.
+                        return new ProbeResult.Success(amount, route);
```
markettes, line 1663: Is it possible to try all and only return the one with lower fees? Are the fees something we could see with the route?

src/Services/LightningService.cs
```diff
@@ -1528,5 +1571,186 @@ public async Task<Dictionary<ulong, ChannelState>> GetChannelsState()
...
+
+                    // Mid-route failure - skip this route and try the next.
+                }
+
+                amount = (long)(amount * ratio);
+            }
+
+            return new ProbeResult.NoRoute("Probe exhausted before reaching minimum amount");
+        }
+        public async Task<Payment> SendPaymentV2Async(Node node, string paymentRequest, long feeLimitMsat,
```
markettes, line 1674:
```suggestion
        }

        public async Task<Payment> SendPaymentV2Async(Node node, string paymentRequest, long feeLimitMsat,
```

src/Services/LightningService.cs
```diff
@@ -1528,5 +1571,186 @@ public async Task<Dictionary<ulong, ChannelState>> GetChannelsState()
...
+            var request = new SendPaymentRequest
+            {
+                PaymentRequest = paymentRequest,
+                FeeLimitMsat = feeLimitMsat,
+                TimeoutSeconds = timeoutSeconds > 0 ? timeoutSeconds : 60,
+                AllowSelfPayment = true,            // mandatory for circular self-pay
+                NoInflightUpdates = true,           // we only care about the terminal Payment update
+            };
+
+            if (outgoingChanIds is { Length: > 0 })
```
markettes, line 1689: :O

src/Services/LightningService.cs
Jossec101: We dont use the route to pay, it's just probing to see it's possible but LND on SendPaymentV2 will optimize this afaik, so the cheaper might be used.

markettes (approved):

src/Services/LightningService.cs
Jossec101: fixed

PR #516, opened 2026-05-15
Change fee settings from UI

markettes:
Enhance the UI for changing fee settings.


Jossec101 (changes requested):
Please introduce validation so:

- Inbound base and fee rate are negative only
- In abs terms, both inbound base and fee rate cannot be bigger than it's counterparter (you cannot surcharge positively a routing, only discount)
- The ui shall require negatives values
- Introduce a default for CLTV (40 blocks its ok), add to constants pls
- Also modify channel request view to set fee rates at creation https://lightning.engineering/api-docs/api/lnd/lightning/open-channel/#lnrpcopenchannelrequest

Jossec101:
> Please introduce validation so:
> 
> * Inbound base and fee rate are negative only
> * In abs terms, both inbound base and fee rate cannot be bigger than it's counterparter (you cannot surcharge positively a routing, only discount)
> * The ui shall require negatives values
> * Introduce a default for CLTV (40 blocks its ok), add to constants pls
> * Also modify channel request view to set fee rates at creation https://lightning.engineering/api-docs/api/lnd/lightning/open-channel/#lnrpcopenchannelrequest

@markettes I guess you are still on this

markettes:
> > Please introduce validation so:
> > 
> > * Inbound base and fee rate are negative only
> > * In abs terms, both inbound base and fee rate cannot be bigger than it's counterparter (you cannot surcharge positively a routing, only discount)
> > * The ui shall require negatives values
> > * Introduce a default for CLTV (40 blocks its ok), add to constants pls
> > * Also modify channel request view to set fee rates at creation https://lightning.engineering/api-docs/api/lnd/lightning/open-channel/#lnrpcopenchannelrequest
> 
> @markettes I guess you are still on this

Yep, you can check now

src/Migrations/ApplicationDbContextModelSnapshot.cs
```diff
@@ -474,6 +474,9 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                     b.Property<int>("MempoolRecommendedFeesType")
                         .HasColumnType("integer");
 
+                    b.Property<string>("RequestMetadata")
+                        .HasColumnType("text");
+
```
Jossec101: mm why going  `json`? I guess we can go with fixed colums as they are more or less explicit what we need in this case

markettes: I don't want to pollute the model with data that is only useful from channel request to open action. After that, that data is trash

Jossec101: but I dont think having a flexible field is worthy here, why the tradeoff of serialisation/and deserialisation? To me it has value to know the fee rate at time of channel opening ?

if you were to say that this data is removed but its not in any form, so I would avoid flexible fields?

src/Pages/Channels.razor
```diff
@@ -311,116 +311,208 @@
...
+                                            <Validation Validator="ValidateChannelFeeRatePpm">
+                                                <FieldLabel>Fee rate (ppm)</FieldLabel>
+                                                <FieldHelp>Outbound proportional fee in parts per million.</FieldHelp>
+                                                <NumericEdit TValue="long?" Min="0L" Max="@((long)uint.MaxValue)"
+                                                                @bind-Value="_channelFeePolicy.FeeRatePpm">
+                                                    <Feedback>
+                                                        <ValidationError />
+                                                    </Feedback>
+                                                </NumericEdit>
+                                            </Validation>
```
Jossec101, line 659: @markettes can we please add here the current channel fee rate? same for inbound

src/Helpers/Constants.cs
```diff
@@ -354,6 +355,9 @@ static Constants()
         var transactionConfBlocks = GetEnvironmentalVariableOrThrowIfNotTesting("TRANSACTION_CONFIRMATION_MINIMUM_BLOCKS");
         if (transactionConfBlocks != null) TRANSACTION_CONFIRMATION_MINIMUM_BLOCKS = int.Parse(transactionConfBlocks);
 
+        var defaultChannelFeePolicyTimelockDeltaBlocks = Environment.GetEnvironmentVariable("DEFAULT_CHANNEL_FEE_POLICY_TIMELOCK_DELTA_BLOCKS");
+        if (defaultChannelFeePolicyTimelockDeltaBlocks != null) DEFAULT_CHANNEL_FEE_POLICY_TIMELOCK_DELTA_BLOCKS = int.Parse(defaultChannelFeePolicyTimelockDeltaBlocks);
```
Jossec101, line 361: Could also add defaults for fee policies? so we dont rely on LND conf (harder to change)

src/Pages/Channels.razor
```diff
@@ -311,116 +311,208 @@
...
+                                        </Field>
+                                    </Column>
+                                </Row>
+                                <Field>
+                                    <Validation Validator="ValidateChannelFeeTimeLockDelta">
+                                        <FieldLabel>Time lock delta</FieldLabel>
+                                        <FieldHelp>CLTV delta advertised for forwarded HTLCs on this channel.</FieldHelp>
+                                        <NumericEdit TValue="long?" Min="0L" Max="@((long)uint.MaxValue)"
+                                                        @bind-Value="_channelFeePolicy.TimeLockDelta">
+                                            <Feedback>
```
Jossec101, line 669: Can we add to this list in the datagrid also a way to see really quick PPM? both inbound and outbound but also what they charge to us (peer policy?) Perhaps peer policy is a tooltip or something, its dense info i let you figure but I think it adds value

src/Pages/Channels.razor
```diff
@@ -942,39 +1125,141 @@
...
         if ((_currentLiquidityRule.MinimumLocalBalance == 0 || _currentLiquidityRule.MinimumLocalBalance == null)
             && (_currentLiquidityRule.MinimumRemoteBalance == 0 || _currentLiquidityRule.MinimumRemoteBalance == null))
         {
             arg1.Status = ValidationStatus.Error;
             arg1.ErrorText = "Minimum local balance cannot be 0 if the minimum remote balance is 0";
         }
 
-    //If the value is 0 is valid
+        //If the value is 0 is valid
         if (_currentLiquidityRule.MinimumRemoteBalance == 0 || _currentLiquidityRule.MinimumRemoteBalance == null)
```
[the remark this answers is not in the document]
markettes: done

src/Migrations/ApplicationDbContextModelSnapshot.cs
markettes: fixed!

src/Helpers/Constants.cs
markettes: wdyt about setting this at node level?

src/Pages/Channels.razor
markettes: Done 🏁

src/Helpers/Constants.cs
markettes: Done 🏁 
Maybe the idea of setting the default fees at node level in NG can be implemented in the future

src/Pages/Channels.razor
markettes: Done!

Jossec101 (approved):

PR #523, opened 2026-06-29
Fix FailureString on HTLC events longer than 1024 chars

imclvr:
Fix for spotted noisy errors like https://app.datadoghq.eu/logs?query=cluster_name%3Aelenpay-production%20-service%3Atraefik%20service%3Anodeguard&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&event=AwAAAZ8KXRwwXBWr2wAAABhBWjhLWFRCQUFBQThuanhiZkZJXzRBQUMAAABdMTE5ZjBhYWEtZTVmMC00OWI3LWIxNDMtODI1MzRhOGE2ODM3LXN5bnRoZXRpYy10aW1lLTE3ODI1ODMyMDAwMDAwMDAwMDBjLTE3ODI1OTAzOTk5ODQwMDAwMDFvAAL_iQ&messageDisplay=inline&refresh_mode=paused&saved-view-id=219284&storage=hot&stream_sort=desc&viz=stream&from_ts=1782584400000&to_ts=1782585600000&live=false

### DIFF
--- a/src/Data/Models/ForwardingHtlcEvent.cs
+++ b/src/Data/Models/ForwardingHtlcEvent.cs
@@ -71,7 +71,7 @@ public class ForwardingHtlcEvent
     public int? WireFailureCode { get; set; }
     public int? FailureDetail { get; set; }
 
-    [MaxLength(1024)]
+    [MaxLength(2048)]
     public string? FailureString { get; set; }
 }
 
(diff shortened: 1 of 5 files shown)


manumonti (approved):
LGTM! 👨🏻‍🔧

src/Data/Models/ForwardingHtlcEvent.cs
```diff
@@ -71,7 +71,7 @@ public class ForwardingHtlcEvent
     public int? WireFailureCode { get; set; }
     public int? FailureDetail { get; set; }
 
-    [MaxLength(1024)]
+    [MaxLength(2048)]
```
RodriFS, line 74: why not leave it at 1024 and just truncate it?

imclvr: I'd prefer to give it the chance to log the whole thing unless extremely big. Take into account that, If most len(FailureStrings) are < 1024, the impact on DB size won't be big. Even if these longer failures are between 1024 and 2048, considering the underneath data structure is varchar.

RodriFS (approved):

Jossec101 (approved):

PR #529, opened 2026-07-02
[GEN-1588] Avoid 0 sats/vb

imclvr:


src/Helpers/Constants.cs
```diff
@@ -119,6 +119,9 @@ public class Constants
     /// </summary>
     public static decimal MAXIMUM_SWAP_OUT_SIZE_BTC = 0.5m;
 
+    //Sat/vb
+    public static decimal MIN_SAT_PER_VB = 1m;
```
Jossec101: why not 0.1? Nowadays we are on sub 1sat/vb level era :)

Jossec101: However its true that bitcoinds need to be configured to relay this

RodriFS: i don't think this is the right approach, i think we should accept an optional minimum amount through grpc. So e.g. you would send something like:

```
mempool_fee_rate = FASTEST_FEE
min_sat_per_vb = 1m
```

imclvr: Will try it out. Anyhow, I approached the hardcoded constant way given we wanted to bring a quick bugfix as what we agreed on.

imclvr: Another approach I prefer, so that we don't affect contracts between components: Should the fee rate obtained is below recommended ones by mempool, we leverage it to the economic fee. 

That way we don't depend on btcd configuration, don't have to modify rpc calls and we don't hardcode values. 

WDYT?

imclvr: @RodriFS @Jossec101 I changed the approach here. 

Turns out, we were always assigning the custom fee on wd request even if the request wanted the "automated" fee from Mempool. That's probably the reason why we have some 0 sat/vb with half hour fees.  Fix is, in such case, assign the fee based on the recommended ones from mempool.

If there's still a 0 for whatever evil reason, let's log it to understand in which cases it's happening.

src/Rpc/NodeGuardService.cs
```diff
@@ -266,6 +266,30 @@ public override async Task<RequestWithdrawalResponse> RequestWithdrawal(RequestW
...
+            {
+                feeRate = request.CustomFeeRate;
+            }
+            else
+            {
+                var recommendedFeeRate = await _nbXplorerService.GetFeesByType(mempoolFeesType);   
+                if (recommendedFeeRate == null)
+                {
+                    _logger.LogWarning("Error getting recommended fee rate for type {feeType}, using 0", mempoolFeesType);
+                    feeRate = 0;
```
Jossec101, line 285: If mempool is down use 1 (?)

imclvr: 1, or 0.1? 🤔 

1 => More money wasted, faster mining.
0.1 => Conservative, slower mining.

RodriFS: mm this kind of breaks this code inside `GenerateTemplatePSBT`
```c#
//We get the mempool.space recommended fees, the custom or use the bitcoin core (nbxplorer) one if not available
var feerate = await _nbXplorerService.GetFeesByType(walletWithdrawalRequest.MempoolRecommendedFeesType)
    ?? walletWithdrawalRequest.CustomFeeRate
    ?? (await LightningHelper.GetFeeRateResult(network, _nbXplorerService)).FeeRate
    .SatoshiPerByte;
```

if the mempool is not working, then we default to `LightningHelper.GetFeeRateResult` as a backup

imclvr: but @RodriFS , for what I saw, `GetFeeRateResult` relies on Mempool for non reg-test, so in this particular case (and rare) where mempool service is not up, it will call it again 🤔

imclvr:
Closing, seems that this case is not being reproduced anymore and needs more consensus

PR #538, opened 2026-07-24
TransferFundsModal: Coin selection, fee selection

RodriFS:
This PR removes the inlined modal for transferring funds into it's own file. Then adds new functionality to it so it has the same features as the Withdrawal modal. I will create a following PR so we can use this file as the common modal for both views (will add an option to transfer to a random address or select a wallet, didn't want to add it here where is not necessary yet)

Screenshot 2026-07-24 at 12 44 52

Stacked PRs
- https://github.com/Elenpay/NodeGuard/pull/539
- -> https://github.com/Elenpay/NodeGuard/pull/538

### DIFF
--- a/src/Shared/TransferFundsModal.razor
+++ b/src/Shared/TransferFundsModal.razor
@@ -1,0 +1,694 @@
+@using System.Text.Json
+@using Blazorise.Extensions
+@using NBXplorer.DerivationStrategy
+@using NBXplorer.Models
+@using NBitcoin
+@using NodeGuard.Jobs
+@using Quartz
+@inject IPriceConversionService PriceConversionService
+@inject IBitcoinService BitcoinService
+@inject IToastService ToastService
+@inject ILightningService LightningService
+@inject IWalletWithdrawalRequestRepository WalletWithdrawalRequestRepository
+@inject IAuditService AuditService
+@inject IWalletWithdrawalRequestPsbtRepository WalletWithdrawalRequestPsbtRepository
+@inject ISchedulerFactory SchedulerFactory
+@inject IWalletRepository WalletRepository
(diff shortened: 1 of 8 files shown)


src/Shared/TransferFundsModal.razor
```diff
@@ -0,0 +1,694 @@
...
+                                            <SelectGroup Label="Mempool.space fee">
+                                                <SelectItem Value="@MempoolRecommendedFeesType.FastestFee">Fastest Fee
+                                                </SelectItem>
+                                                <SelectItem Value="@MempoolRecommendedFeesType.EconomyFee">Economy Fee
+                                                </SelectItem>
+                                                <SelectItem Value="@MempoolRecommendedFeesType.HourFee">Hour Fee
+                                                </SelectItem>
+                                                <SelectItem Value="@MempoolRecommendedFeesType.HalfHourFee">Half Hour
+                                                    Fee
+                                                </SelectItem>
```
imclvr, line 171: recommended fees are broadly used, perhaps would be great to have a fee selector component that can be reused. 

PS: Yes, I know, this sounds too React 😬 , but the more we don't duplicate the better

RodriFS: will do in a following PR

imclvr:
I think this deserves its own bUnit tests.

imclvr (dismissed):

manumonti (dismissed):
LGTM! 🎸

RodriFS:
> I think this deserves its own bUnit tests.

I did my best, but Blazorise is not compatible with bunit 😭. It's impossible to mock.
We can't update blazorize because now the licence change and it's no longer open source

imclvr (approved):

PR #544, opened 2026-07-29
fix: quiet benign duplicate-HTLC EF errors and disable sensitive-data logging in prod

markettes:
Production logs were being flooded with large `Error`-level entries every time a
duplicate `ForwardingHtlcEvent` was inserted (composite PK collision, Postgres `23505`).

These duplicates are **expected and already handled**: `ForwardingHtlcEventRepository.UpsertAsync`
catches the unique-violation and skips the row. But two things made them noisy anyway:

1. **EF Core logs the failure independently of our `catch`.** During the failing command EF
   emits `RelationalEventId.CommandError` and `CoreEventId.SaveChangesFailed` at `Error` — a full
   SQL command + parameter dump and a complete stack trace — before the exception ever reaches
   our handler. Serilog pins `Microsoft` to `Warning`, but these are `Error`, so they passed through.
2. **Sensitive-data logging was on in prod.** `AddDbContextFactory<ApplicationDbContext>` called
   `EnableSensitiveDataLogging()` + `EnableDetailedErrors()` unconditionally, dumping raw parameter
   values (node pubkeys, HTLC amounts, peer aliases) into every failed-command log — both noise
   and a privacy concern.

### DIFF
--- a/src/Data/Repositories/ForwardingHtlcEventRepository.cs
+++ b/src/Data/Repositories/ForwardingHtlcEventRepository.cs
@@ -87,7 +87,11 @@ public class ForwardingHtlcEventRepository : IForwardingHtlcEventRepository
         }
         catch (DbUpdateException e) when (e.InnerException is PostgresException pg && pg.SqlState == PostgresErrorCodes.UniqueViolation)
         {
-            _logger.LogDebug("Skipping duplicated forwarding HTLC event for node {ManagedNodePubKey}", forwardingHtlcEvent.ManagedNodePubKey);
+            _logger.LogDebug(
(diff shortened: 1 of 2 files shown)


Jossec101:
Nack, can't we handle this without removing logging? is the unique really what I thought it should be?

RodriFS (dismissed):

Jossec101 (changes requested):
Nack, see comment above

RodriFS:
> Nack, see comment above

didn't see the comment, sorry

Jossec101:
Closing as this does not happen too much @markettes

PR #547, opened 2026-07-30
Add Peers page for bulk dynamic-fee management

markettes:
Adds a node-scoped /peers page (NodeManager/Superadmin) that groups the selected node's open channels by remote peer.
Lets you enable/disable dynamic fee management for all channels to a peer in one click, instead of channel by channel.
Includes a per-node dynamic-fee gate toggle and light per-peer stats (channel count, capacity, live local/remote split, dynamic-fee count).


src/Shared/NavMenu.razor
```diff
@@ -80,6 +80,14 @@
             </div>
         </AuthorizeView>
 
+        <AuthorizeView Roles="Superadmin,NodeManager">
+            <div class="nav-item px-3">
+                <NavLink class="nav-link" href="peers">
+                    <span class="oi oi-people" aria-hidden="true"></span> Peers
```
Jossec101: I would not name this peers as I  still envision the peer view (like the channels view) to manage rules via rebalancing BUT, perhaps we can name this `Dynamic Fee Mngmt` ? or somethign? Or `Routing`?

markettes: done!

Jossec101 (approved):

PR #556, opened 2026-08-05
Reject 2-of-2 wallets co-signed by internal wallet

Jossec101:
Block creation and finalisation of 2-of-2 multisig wallets where one of the two required keys is NodeGuard's internal co-signing key. Because NodeGuard co-signs automatically, such wallets collapse to single-sig — a lone human keyholder combined with the automatic co-signature is enough to move funds.

- Add `IsEffectivelySingleSigTwoOfTwo` helper and a shared `TwoOfTwoNotAllowedMessage` on `Wallet` for reuse by guards, UI and tests.
- Guard `WalletRepository.AddAsync` (defence in depth for direct/API inserts) and `FinaliseWallet` (the primary UI path, where `Keys` and `MofN` are final).
- Block finalisation in the UI too: the finalise modal explains the rejection and disables its confirm button for such a wallet, `FinaliseWallet` mirrors the repository guard in case the modal state is stale, and the failure toast now surfaces the repository's reason instead of a generic message.
- Exclude watch-only, BIP39 and hot wallets, which NodeGuard does not auto co-sign.

### DIFF
--- a/src/Data/Models/Wallet.cs
+++ b/src/Data/Models/Wallet.cs
@@ -115,6 +115,24 @@ namespace NodeGuard.Data.Models
 
         [NotMapped] public bool RequiresInternalWalletSigning => IsBIP39Imported || IsWatchOnly ? false : Keys != null ? Keys.Count == MofN : false;
 
+        /// <summary>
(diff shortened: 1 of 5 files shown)


markettes (approved):
LGTM

PR #557, opened 2026-08-05
Validate human PSBT approvals against template

Jossec101:
Add PsbtApprovalValidator gate to ChannelOperationRequestPSBT and WalletWithdrawalRequestPSBT repositories to reject malformed or duplicate human signatures before they reach the signing path.

- Wire ValidateApproval into AddAsync/AddRangeAsync for channel operation PSBTs, using SIGHASH_NONE to match ChannelRequests.razor.
- Server-generated rows (template, internal wallet, finalised) bypass validation; human approvals are checked against the request's template PSBT and existing signatures.
- Include ChannelOperationRequestPsbts when loading the request so the validator has the full signature set.
- Log rejections with the request id and reason, and surface the error to the caller.
- Add BitcoinService tests covering the duplicate-template-row and related rejection scenarios on the withdrawal signing path.


test/NodeGuard.Tests/E2E/PsbtApprovalBindingE2ETests.cs
```diff
@@ -0,0 +1,410 @@
...
+    }
+
+    private static NodeGuardService.NodeGuardServiceClient CreateGrpcClient(out Metadata headers)
+    {
+        headers = new Metadata { { "auth-token", Env("NODEGUARD_API_TOKEN", DefaultDevToken) } };
+        var endpoint = Env("NODEGUARD_GRPC_ENDPOINT", "http://localhost:50051");
+        return new NodeGuardService.NodeGuardServiceClient(GrpcChannel.ForAddress(endpoint));
+    }
+
+    private const string DefaultDevToken = "8rvSsUGeyXXdDQrHctcTey/xtHdZQEn945KHwccKp9Q=";
```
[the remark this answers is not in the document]
Jossec101: @AikidoSec ignore: E2E test.

src/Helpers/PsbtApprovalValidator.cs
```diff
@@ -0,0 +1,143 @@
...
+        var submittedOutpoints = submitted.Inputs.Select(x => x.PrevOut).ToHashSet();
+        if (!templateOutpoints.SetEquals(submittedOutpoints))
+        {
+            return Result.Fail("Invalid PSBT, the UTXOs do not match the ones this request was created for.");
+        }
+
+        if (!submitted.Inputs.All(x => x.PartialSigs.Any()))
+        {
+            return Result.Fail($"Invalid PSBT, every input must be signed with Sighash: {expectedSigHash}.");
+        }
```
markettes, line 101: If we want this to be sighash type agnostic, we should check the outpoints too, right?

src/Services/BitcoinService.cs
```diff
@@ -408,6 +408,35 @@ public async Task PerformWithdrawal(WalletWithdrawalRequest walletWithdrawalRequ
...
+                    templatePsbt.GetGlobalTransaction().GetHash())
+                {
+                    var mismatch =
+                        $"The PSBT to be signed for withdrawal request:{walletWithdrawalRequest.Id} does not " +
+                        "match the approved transaction, refusing to sign";
+                    _logger.LogError(mismatch);
+
+                    throw new ArgumentException(mismatch);
+                }
+
```
markettes, line 439: why not reusing `PsbtApprovalValidator` partially?

src/Helpers/PsbtApprovalValidator.cs
Jossec101: but sighashes are on the inputs?

markettes: https://developer.bitcoin.org/devguide/transactions.html#signature-hash-types
anyonecanpay flag yes, without the flag is about the outputs

Jossec101: we dont use that :P

markettes: exactly, we don't use `anyonecanpay` that is for the inputs, we should check the outputs. Am I losing something?

Jossec101: Fixed comment with proper message "It needs to be sign every partial sig"

src/Services/BitcoinService.cs
Jossec101: Extracted func

markettes (approved):

PR #563, opened 2026-08-12
[GEN-1815] Refactor coin selection

imclvr:
This refactoring allows a common and meaningful place to put utxo selection from NG. 

---


imclvr:
The approach is a bit conservative, we can attack directly to nbxplorer and make coin selection to not go through custom algorithms on Nodeguard. Clarified after discussing about how to proceed.

PR #566, opened 2026-08-12
[GEN-1815] Add naive approach to avoid utxo selection to leave no margin for fees

imclvr:
Idea taken by `SelectUTXOsByOldest` approach. UTXOs are selected even after satsAmount is hit by the sum of utxo values. So this will be the stopper for `SelectUTXOsByClosest` as well. Other approaches will imply bigger and potentially riskier changes.

---


imclvr:
The approach is a bit conservative, we can attack directly to nbxplorer and make coin selection to not go through custom algorithms on Nodeguard. Clarified after discussing about how to proceed.

PR #570, opened 2026-08-17
Add feature to change maximum HTLC value

markettes:
This pull request adds support for configuring the maximum HTLC (Hash Time Locked Contract) value per channel in the application, allowing users to view and update the maximum HTLC (in satoshis) that a channel can forward. The changes include updates to the UI, validation logic, backend service interfaces, and tests to support this new functionality.

### DIFF
--- a/src/Pages/Channels.razor
+++ b/src/Pages/Channels.razor
@@ -566,6 +566,10 @@
                                                                     <td class="fw-bold">Time Lock Delta:</td>
                                                                     <td><Badge Color="Color.Light">@_currentFeePolicies.ManagedNodePolicy.TimeLockDelta</Badge></td>
                                                                 </tr>
+                                                                <tr>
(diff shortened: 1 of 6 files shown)


src/Pages/Channels.razor
```diff
@@ -566,6 +566,10 @@
                                                                     <td class="fw-bold">Time Lock Delta:</td>
                                                                     <td><Badge Color="Color.Light">@_currentFeePolicies.ManagedNodePolicy.TimeLockDelta</Badge></td>
                                                                 </tr>
+                                                                <tr>
```
Jossec101, line 569: I dont think we need to expose this to users

Jossec101 (changes requested):
Why not make it a constant/envvar? I dont see us changing this value or at least being per node? When NG first starts we simply put call updatechanpolicy on all nodes with this value?

PR #575, opened 2026-08-24
Scope coin-selection ignore list to queried wallet

Jossec101:
Refine the coin-selection process by ensuring the ignore list only includes locked and frozen UTXOs belonging to the queried wallet. 

This change prevents unnecessary outpoints from other wallets from affecting the request size, which previously led to HTTP 414 errors. Additionally, optimize the handling of dust UTXOs by allowing the backend to filter them by value, reducing the number of query parameters needed.

 Implement tests to cover the new functionality and ensure compatibility with existing systems.

### DIFF
--- a/src/Services/NBXplorerService.cs
+++ b/src/Services/NBXplorerService.cs
@@ -162,16 +162,26 @@ public class NBXplorerService : INBXplorerService
                 new("strategy", strategy.ToString()),
                 new("limit", limit.ToString()),
                 new("amount", amount.ToString()),
+                // Lets the backend drop dust by value, so callers do not have to spend one query
+                // parameter per dust outpoint saying the same thing
+                new("minimumValue", Constants.MINIMUM_UTXO_VALUE_SATS.ToString()),
             };
             if (strategy == CoinSelectionStrategy.ClosestToTargetFirst)
             {
                 keyValuePairs.Add(new("closestTo", closestTo.ToString()));
             }
 
-            ignoreOutpoints?.ForEach(outpoint => keyValuePairs.Add(new("ignoreOutpoint", outpoint)));
-
             var url = QueryHelpers.AddQueryString(requestUri, keyValuePairs);
-            var response = await _httpClient.GetAsync(url, cancellation);
+
+            // The ignored outpoints travel in the body. As one repeated ignoreOutpoint query
+            // parameter each they cost ~83 bytes apiece, so about ninety of them overflowed
+            // Kestrel's 8KB MaxRequestLineSize and NBXplorer answered 414 — which the callers
+            // swallow into a degraded selection. The body limit is 30MB, roughly 3,662x the
+            // room, so the list no longer has a practical ceiling.
+            // NOTE: this requires an NBXplorer serving POST /selectutxos (Elenpay/NBXplorer#18).
+            // Against an older build every call answers 405 and coin selection degrades.
+            var response = await _httpClient.PostAsync(url,
+                JsonContent.Create(new { ignoreOutpoints }), cancellation);
 
             if (response.IsSuccessStatusCode)
             {
--- a/src/Services/CoinSelectionService.cs
+++ b/src/Services/CoinSelectionService.cs
@@ -160,16 +160,28 @@ public class CoinSelectionService: ICoinSelectionService
     }
 
     /// <summary>
-    /// Outpoints that must never be offered for coin selection: locked, frozen and dust UTXOs.
+    /// Outpoints that must never be offered for coin selection: the locked and frozen UTXOs that
+    /// belong to this wallet.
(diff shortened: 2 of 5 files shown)


src/Services/NBXplorerService.cs
```diff
@@ -162,16 +174,25 @@ public async Task<UTXOChanges> GetUTXOsByLimitAsync(DerivationStrategyBase deriv
...
+            var url = QueryHelpers.AddQueryString(urlWithoutOutpoints,
+                ignoreOutpoints?.Select(outpoint =>
+                    new KeyValuePair<string, string?>("ignoreOutpoint", outpoint))
+                ?? Enumerable.Empty<KeyValuePair<string, string?>>());
 
-            var url = QueryHelpers.AddQueryString(requestUri, keyValuePairs);
-            var response = await _httpClient.GetAsync(url, cancellation);
+            var response = url.Length <= MaxSelectUtxosUrlLength
+                ? await _httpClient.GetAsync(url, cancellation)
+                : await _httpClient.PostAsync(urlWithoutOutpoints,
```
imclvr: I believe the body will also have a size constrain, probably much bigger than http request and thus we're gonna be safe, right? 

And the chance of having a huge body that cannot be swallowed by nbxplorer is 1E-29732837283, right?

imclvr (dismissed):

src/Services/NBXplorerService.cs
```diff
@@ -162,16 +174,25 @@ public async Task<UTXOChanges> GetUTXOsByLimitAsync(DerivationStrategyBase deriv
...
+                    JsonContent.Create(new { ignoreOutpoints }), cancellation);
```
RodriFS: why not just use POST always?

RodriFS (dismissed):

src/Services/NBXplorerService.cs
Jossec101: Yes on both counts, and I checked the numbers rather than trusting my memory of them.

Instantiating `KestrelServerLimits` on net10.0 gives:

```
MaxRequestLineSize = 8192
MaxRequestBodySize = 30000000
```

So the body gets ~3,662x the room the request line does. An outpoint costs ~69 bytes as a JSON array element (`"<64 hex>-<n>",`), which puts the body ceiling around **430,000 outpoints**, versus ~98 on the request line. A wallet would need 430k locked-or-frozen UTXOs in a single wallet to reach it, so agreed: your exponent is about right.

Worth adding that it is not a new failure mode even in that impossible case. Blowing the body limit is a 413, which lands in the same `catch` as everything else: the Blazor path falls back to the plain UTXO listing and the local filter still strips locked/frozen/dust, and the gRPC path returns an empty selection exactly as it does today. So the worst case degrades the same way it already did before this PR, just 3,662x further out.

src/Services/NBXplorerService.cs
Jossec101: Rollout safety — it is deliberately temporary, not a permanent design.

The POST route only exists in Elenpay/NBXplorer#18, which is approved but not merged, so it is not in a published image yet. `docker/docker-compose.dev.yml` pins the floating `ghcr.io/elenpay/nbxplorer:elenpay-develop` tag with no `pull_policy`, so even after that merges, any environment holding a cached copy keeps serving the GET-only route until it explicitly re-pulls.

If we always POSTed, then against any NBXplorer that predates #18 **every** `selectutxos` call 405s. That gets swallowed into the same degraded path as a 414, so coin selection would silently stop applying strategy/limit/amount for *all* wallets — not just the ones with a long ignore list. We would have widened the exact bug this PR is fixing, and only while both repos are mid-rollout, which is the worst time for it.

The length branch confines the dependency on the new route to the requests that are already broken today. Short lists, which is nearly all of them, keep working against any backend, so NodeGuard and NBXplorer can ship in either order without a coordinated deploy — and the compose pin being a mutable tag means "the new image is everywhere" is not something we can actually verify.

Agreed it should not live forever, though. Once #18 is merged and the pin is moved to an immutable tag of a build carrying the route, the GET branch and `MaxSelectUtxosUrlLength` can both go and this collapses to an unconditional POST. Happy to add a TODO in the code pointing at that, or to do it as a follow-up PR once the pin is bumped — the pin bump is already queued as the next PR in this stack.

RodriFS (approved):

src/Services/NBXplorerService.cs
Jossec101: Clankers

Jossec101: Fixed

src/Services/NBXplorerService.cs
Jossec101: Anyway, post is enforced.

PR #576, opened 2026-08-24
E2E: coin selection with an ignore list too long for the request line

Jossec101:
Guards the transport switch end to end. The test funds a wallet with 130 small
outputs plus a few larger ones in a single transaction, freezes only the small
ones through the AddTags RPC, and asserts GetAvailableUtxos still returns the
wallet's real UTXOs and never surfaces a frozen one.

Frozen UTXOs rather than dust carry the list on purpose. Dust is excluded by
value now, so a dust-driven test would stay green even if the exclusion list
never reached the server; frozen UTXOs above the dust floor have no other
reason to be missing. The test also thaws one at the end and asserts it comes
back, which pins their absence on the tag rather than on anything incidental.

Before the fix this is a 414 that GetAvailableUtxos swallows into an empty
UTXOChanges, so the RPC reports success with nothing selectable for a 20 BTC
wallet. That is the assertion that flips.

Uses the fourth seeded wallet, not the hot wallet the rest of the suite
withdraws from, because the frozen outputs stay in it for the rest of the run.

The suite needs an NBXplorer serving POST selectutxos, so the runner gets an
NBXPLORER_URI and the test probes the route first, failing with that diagnosis
rather than looking like a coin-selection bug. Without the probe it would be
red identically before and after the fix, which is no test at all.

TODO before this is merged: repin the nbxplorer image to the immutable tag of a
build carrying the POST route. See the comment in docker/docker-compose.dev.yml.

---

### DIFF
--- a/docker/e2e/README.md
+++ b/docker/e2e/README.md
@@ -49,6 +49,13 @@ Tests live in `test/NodeGuard.Tests/E2E/`. The rebalance/fee-engine scenarios ar
   `{NODE}_HOST` + `{NODE}_MACAROON` (from `extract-env`). `LndTestClient` reads them from the process env
   or, failing that, straight from `nodeguard-macaroons.env` on the mounted `e2e_env` volume — so it works
   without the runner entrypoint exporting them.
+- **NBXplorer must serve `POST selectutxos`**: NodeGuard sends the coin-selection exclusion list in a
+  request body, so a GET-only NBXplorer answers 405 to every `selectutxos` call and coin selection degrades
+  for every wallet, not just the ones this test covers. It probes `NBXPLORER_URI` first
+  (`Allow: GET` = pre-fix, `Allow: GET, POST` = patched) and fails with that diagnostic rather than looking
+  like a coin-selection bug — without the probe the test is red identically pre-fix and post-fix, because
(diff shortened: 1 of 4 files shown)


docker/e2e/README.md
```diff
@@ -49,6 +49,12 @@ Tests live in `test/NodeGuard.Tests/E2E/`. The rebalance/fee-engine scenarios ar
   `{NODE}_HOST` + `{NODE}_MACAROON` (from `extract-env`). `LndTestClient` reads them from the process env
   or, failing that, straight from `nodeguard-macaroons.env` on the mounted `e2e_env` volume — so it works
   without the runner entrypoint exporting them.
+- **NBXplorer must serve `POST selectutxos`**: `LargeIgnoreListCoinSelectionE2ETests` exercises the transport
+  NodeGuard switches to once the exclusion list outgrows the request line. It probes `NBXPLORER_URI` first
+  (`Allow: GET` = pre-fix, `Allow: GET, POST` = patched) and fails with that diagnostic rather than looking
```
RodriFS: i think this is now outdated

Jossec101: Done

RodriFS (approved):

PR #580, opened 2026-08-26
Fix sweep node wallets job to return errors without clear reason

imclvr:
```
        _logger.LogError("Error while funding sweep transaction reason: {Reason}", reason)
```
        
There're many error traces in prod without a clear reason. After looking at the code, looks like we were missing one particular case which seems to be a quite frequent one.

### DIFF
--- a/src/Jobs/SweepNodeWalletsJob.cs
+++ b/src/Jobs/SweepNodeWalletsJob.cs
@@ -95,69 +95,67 @@ public class SweepNodeWalletsJob : IJob
                     });
                 var totalSatsAvailable = utxos.Sum(x => x.AmountSat);
 
-                if (returningAddress != null && lndChangeAddress != null && utxos.Any() && totalSatsAvailable > Constants.MINIMUM_SWEEP_TRANSACTION_AMOUNT_SATS)
+                var canSweep = returningAddress != null && lndChangeAddress != null && utxos.Any() && totalSatsAvailable > Constants.MINIMUM_SWEEP_TRANSACTION_AMOUNT_SATS;
+
+                if (!canSweep)
                 {
-                    // We need to maintain onchain balance to be at least RequiredAnchorChannelClosingAmount but also we apply a 10% buffer to pay for this sweep fees and let some more money on the wallet
-                    var sweepedFundsAmount = (long)((totalSatsAvailable - requiredAnchorChannelClosingAmount) * 0.9); 
-                    var sendManyResponse = await lightningClient.SendManyAsync(new SendManyRequest()
+                    // Sweep can't be performed, log the reason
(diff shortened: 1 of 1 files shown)


src/Jobs/SweepNodeWalletsJob.cs
```diff
@@ -266,4 +254,26 @@ await _auditService.LogSystemAsync(
         }
         _logger.LogInformation("{JobName} ended on node: {NodeName}", nameof(SweepNodeWalletsJob), node.Name);
     }
+
+    private static string GetReasonFailureSweeping(List<Utxo> utxos, long requiredAnchorChannelClosingAmount, NBXplorer.Models.KeyPathInformation? returningAddress, NewAddressResponse? lndChangeAddress, long totalSatsAvailable)
```
RodriFS: maybe we can return (bool, string). bool meaing "true" if to show an error and "false" we show a warning?
Some of this "errors" are expected so there is no point in setting the level to error

imclvr: Fair enough, or just a GetReasonFailure returning empty and, if so, fallbacking to `GetReasonWarning` for example.

manumonti (dismissed):
LGTM!!! 🙌

src/Jobs/SweepNodeWalletsJob.cs
```diff
@@ -266,4 +264,34 @@ await _auditService.LogSystemAsync(
...
+        {
+            return "Returning address not found / null";
+        }
+        else if (lndChangeAddress == null)
+        {
+            return "LND returning address not found / null";
+        }
+
+        return string.Empty;
+    }
```
RodriFS, line 282: nit: null is more explicit
```suggestion
    private static string? GetReasonFailureErrorSweeping(
        NBXplorer.Models.KeyPathInformation? returningAddress, NewAddressResponse? lndChangeAddress
    )
    {
        if (returningAddress == null)
        {
            return "Returning address not found / null";
        }
        else if (lndChangeAddress == null)
        {
            return "LND returning address not found / null";
        }

        return null;
    }
```

RodriFS (approved):

src/Jobs/SweepNodeWalletsJob.cs
imclvr: I try to follow the approach from this response: https://stackoverflow.com/questions/11730339/defining-strings-as-null-is-better-or-as-empty-strings (unless no way to, or there is a clear distinction between "" and a non-existent thing), as  nulls are prone to errors in prod.

For example, null to me could match to a difference between 0 (there is nothing) and this does not even exist.

RodriFS: makes sense!

PR #583, opened 2026-08-27
Create Auto Rebalance Algorithm

markettes:
Adds automated channel rebalancing to the routing engine: `RebalanceInitiatorService` (pure,
unit-tested) sorts channels into drainable sources and depleted destination peers and pairs them
into sized, profit-gated plans, which `AutoRebalanceJob` dispatches through `RebalanceService` on
its own 10-minute cadence.

Every plan is bounded before it goes out — per-channel `IsAutoRebalanceEnabled` opt-in, a deadband
either side of each channel's target ratio, a fee ceiling of `CostToEarnRatio × the destination
peer's earn rate`, and per-node budget / in-flight / per-run caps — plus per-node dry-run and the
global `ROUTING_ENGINE_ENABLED` kill switch, which is off by default, so merging this changes no
behaviour until it's switched on.

Supporting changes: `ChannelRoutingState` becomes per (channel, managed node) so both ends of a
channel NodeGuard manages actuate independently; the new `RoutingEngineSnapshotService` builds the
per-node channel view now shared with `ChannelFeeOptimizerJob`; two migrations and unit tests for
the planner and the job.

### DIFF
--- a/src/Helpers/Constants.cs
+++ b/src/Helpers/Constants.cs
@@ -284,6 +284,12 @@ public class Constants
     /// </summary>
     public static int ROUTING_ENGINE_ACTUATOR_OFFSET_MINUTES = 5;
 
+    /// <summary>
+    /// Cadence of AutoRebalanceJob in prod, in minutes. Independent of
+    /// ROUTING_ENGINE_JOB_INTERVAL_MINUTES. In dev it is 1 minute.
+    /// </summary>
+    public static int ROUTING_ENGINE_REBALANCE_JOB_INTERVAL_MINUTES = 10;
+
     // Both fees use integral control: each cycle the applied value is nudged by gain·deviation·baseline
     // off its previous value, so a persistent deviation keeps driving the fee until the channel balances.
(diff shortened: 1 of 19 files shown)


src/Helpers/Constants.cs
```diff
@@ -288,7 +300,7 @@ public class Constants
     public static double ROUTING_ENGINE_FEE_DEADBAND = 0.03;
 
     // The rebalancer's imbalance deadband is a separate, more aggressive threshold for triggering rebalances.
-    public static double ROUTING_ENGINE_REBALANCE_TRIGGER = 0.15;
+    public static double ROUTING_ENGINE_REBALANCE_DEADBAND = 0.15;
```
Jossec101, line 303: Can you doc this better?

Jossec101 (dismissed):

Jossec101 (approved):

src/Helpers/Constants.cs
Jossec101: Resolve if done in a follow up

PR #587, opened 2026-09-01
Add pessimistic fee calculation in RebalanceRepository and set timeout

markettes:
Implement a new method for pessimistic fee consumption in the RebalanceRepository and set a timeout for fee calculations in the AutoRebalanceJob. These changes enhance the accuracy of fee tracking and improve the handling of rebalance operations.


src/Data/Repositories/RebalanceRepository.cs
```diff
@@ -166,6 +166,33 @@ public async Task<long> GetConsumedFeesSince(int nodeId, DateTimeOffset since)
...
+            var feePaid = r.FeePaidSats ?? 0;
+            // Conservative worst-case reservation for a still-in-flight rebalance.
+            var reserved = Rebalance.WorstCaseFeeSats(r.RequestedAmountSats, r.MaxFeePct);
+            total += r.Status == RebalanceStatus.Succeeded
+                ? feePaid
+                : reserved;
+        }
+
+        return total;
+    }
```
Jossec101: Can't this sum be done on SQL-level for perf?

src/Helpers/Constants.cs
```diff
@@ -368,6 +368,10 @@ public class Constants
     /// RebalanceBudgetRefreshInterval unset.
     public static int ROUTING_ENGINE_REBALANCE_DEFAULT_BUDGET_REFRESH_HOURS = 24;
 
+    /// Payment timeout (seconds) handed to each rebalance the routing-engine actuator initiates.
+    /// Circular routes need more room than the RebalanceRequest default of 60s.
+    public static int ROUTING_ENGINE_REBALANCE_TIMEOUT_SECONDS = 180;
```
Jossec101: I do think we can move this to rebalance config in general (manual and auto)

src/Data/Repositories/RebalanceRepository.cs
markettes: done, mb

src/Helpers/Constants.cs
markettes: done

Jossec101: we discussed to move this to a rebalancing feature pure config (env var constant)

Jossec101 (approved):

PR #591, opened 2026-09-03
[GEN-1615] Enable settlement coin selection by 'closest'  strategy from NBXplorer

imclvr:
for coin settlement, we are disregarding old, manual selection of utxos by oldest strategy. 

We want to start reusing nbxplorer selection by strategy features for that, avoiding duplication. And for this case we are going to try a different strategy, selecting coins by closest instead of oldest, as an attempt to try reducing dust.


RodriFS:
In ep-backend we do `GetAvailableUtxos` with strategy `COIN_SELECTION_STRATEGY_CLOSEST_TO_TARGET_FIRST` for withdrawals and then pass the utxos to the withdrawal request call. So hardcoding it to the path where no utxos are selected won't do anything for the backend.
Using an environmental variable to hardcode a selection strategy is also something that i think should be reconsidered in my opinion

PR #594, opened 2026-09-04
Add RBF fee bump support via gRPC with error handling and e2e tests

Jossec101:
## Why

PR #557 removed a redundant, untagged PSBT row that the Withdrawals page stored after generating a hot-wallet template. That row's insert had a hidden side effect in `WalletWithdrawalRequestPsbtRepository.AddAsync`: it moved the request to `PSBTSignaturesPending`, the only status `PerformWithdrawal` accepts. Without it, hot-wallet requests created from the page (fresh withdrawals from the datagrid and every RBF fee bump) stayed `Pending`, `PerformWithdrawalJob` threw "Invalid status", marked the bump `Failed` and reset the original, while the UI reported success. Transfer Funds and the gRPC `RequestWithdrawal` set the status explicitly, which is why the gRPC-driven E2E suite never caught it.

## What

- **`WithdrawalRequestService`** (new): owns the RBF replacement flow — fee bumps and cancellations — (validation, creation with the original's UTXOs locked, hot wallets created straight in `PSBTSignaturesPending`) and the hot-wallet execution step (promote status, template, mark the original `Bumped` / `Cancelled`, schedule `PerformWithdrawalJob`, revert on failure). `Withdrawals.razor` and `BumpfeeModal.razor` call it; the page no longer overwrites a replacement's fields with stale form state.
- **`BumpWithdrawal` gRPC method**: same orchestration over the API. The withdrawal is addressed by request id **or by the txid `RequestWithdrawal` returned** (`oneof target`). Refusals carry a typed `BumpingErrorReason`, mapped to `NotFound` / `FailedPrecondition` / `InvalidArgument`.
- **Cancel via RBF** (`CancelWithdrawal` gRPC method + "Cancel (RBF)" action next to "Bump fee" in the UI): replaces the unconfirmed transaction with one that spends the same inputs at a higher fee rate and returns everything to a fresh address of the wallet, so the destination is never paid. The original ends up `Cancelled` with a reason pointing at the replacement (restored to pending if the replacement is abandoned). New `IsRbfCancellation` column (migration `AddIsRbfCancellationToWalletWithdrawalRequest`) so the cold-wallet approval path and the failure paths treat the original correctly. Replacement requests skip the wallet-balance check on creation, since they re-spend inputs that are already locked.
- **Tests**: `WithdrawalRequestServiceTests` (bump/cancel creation, status promotion ordering, revert paths), `BumpWithdrawal` / `CancelWithdrawal` cases in `NodeGuardServiceTests` (hot/cold, error mapping, txid resolution), and `WithdrawalRbfBumpE2ETests` with two scenarios: bump by txid (same inputs and destination, original evicted and `WITHDRAWAL_BUMPED`, replacement settled once mined) and cancel by txid (single output back to the wallet, destination never paid, original `WITHDRAWAL_CANCELLED`, funds a spendable UTXO of the wallet again).
- E2E stack: `MONITOR_WITHDRAWALS_CRON` set to 10 s so settlement is observable.

## Verification

- `dotnet test --filter "Category!=E2E"`: 532 green.
- CI E2E (`run-e2e-tests`, [run 33898507472](https://github.com/Elenpay/NodeGuard/actions/runs/33898507472)): the whole suite green (23 tests), including both RBF scenarios — `BumpWithdrawal_ReplacesTheUnconfirmedHotWalletTransaction` (bump by txid) and `CancelWithdrawal_ReturnsTheFundsOfTheUnconfirmedHotWalletTransaction`.
- Note for local runs: the containerized stack needs an NBXplorer image with the `POST selectutxos` route.

### DIFF
--- a/src/Rpc/NodeGuardService.cs
+++ b/src/Rpc/NodeGuardService.cs
@@ -271,7 +271,7 @@ public class NodeGuardService : Nodeguard.NodeGuardService.NodeGuardServiceBase,
                     throw new RpcException(new Status(StatusCode.Internal, "Derivation strategy not found"));
 
                 utxos = await _coinSelectionService.GetUTXOsByOutpointAsync(derivationStrategyBase, outpoints);
-      
+
             }
 
             // Create destination objects for the withdrawal request
@@ -391,12 +391,14 @@ public class NodeGuardService : Nodeguard.NodeGuardService.NodeGuardServiceBase,
         ServerCallContext context)
     {
         WalletWithdrawalRequest? bumpRequest = null;
+        var target = DescribeBumpTarget(request);
         try
         {
             var feeType = (MempoolRecommendedFeesType)request.MempoolFeeRate;
             decimal? customFeeRate = request.HasCustomFeeRate ? request.CustomFeeRate : null;
 
-            bumpRequest = await _withdrawalRequestService.CreateBumpRequestAsync(request.RequestId, feeType, customFeeRate);
+            var originalRequestId = await ResolveBumpTargetAsync(request);
+            bumpRequest = await _withdrawalRequestService.CreateBumpRequestAsync(originalRequestId, feeType, customFeeRate);
 
             var isHotWallet = bumpRequest.Wallet?.IsHotWallet ?? false;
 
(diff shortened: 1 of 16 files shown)


RodriFS (dismissed):

src/Rpc/NodeGuardService.cs
```diff
@@ -382,6 +389,152 @@ private void CancelWithdrawalRequest(WalletWithdrawalRequest? withdrawalRequest)
...
+
+        var (requestId, txid, isHotWallet) = await ReplaceWithdrawalAsync(target,
+            (MempoolRecommendedFeesType)request.MempoolFeeRate,
+            request.HasCustomFeeRate ? request.CustomFeeRate : null,
+            cancellation: false);
+
+        return new BumpWithdrawalResponse { RequestId = requestId, Txid = txid, IsHotWallet = isHotWallet };
+    }
+
+    public override async Task<CancelWithdrawalResponse> CancelWithdrawal(CancelWithdrawalRequest request,
```
RodriFS, line 410: should this be audited?

Jossec101: I do think it's sound, it's a normal RBF where outputs go to our wallet (original), e2e tested and UI-tested https://github.com/Elenpay/NodeGuard/blob/5057dc7d1e1553e0933f92cab9346f2722b1d329/test/NodeGuard.Tests/E2E/WithdrawalRbfBumpE2ETests.cs#L159

What do you see?

Jossec101: Perhaps a tweak is to return to the same input address for being more conservative though that's a derivation path issue.

RodriFS (approved):

PR #595, opened 2026-09-07
Add rebalance count and activity panels to the routing dashboard

markettes:
Adds two panels to the Routing dashboard, in a new row under "% Routed of total":

- **Rebalances (total)** — every rebalance created in the window on the selected nodes. One row per
  rebalance (retries reuse the row and bump `AttemptNumber`), so this counts rebalances, not attempts.
- **Rebalances over time** — stacked bars per interval, split by outcome: Succeeded / Failed / In flight.
  "Failed" is any status that is neither settled nor still running, so the three series always sum to the
  total in the stat beside it.


docker/grafana/provisioning/dashboards/routing.json
Jossec101, line 1: I would add a success rate over and one over timee? trying to condense in 1 widget as possible?

markettes: I have added rebalance amount over time, I didn't see much value in success rate for the rebalances, wdyt?

Jossec101: I do, to detect/see a trend of rebalancing going up/down, trends are important

markettes: done!

Jossec101 (approved):

PR #597, opened 2026-09-07
Enforce single template PSBT per request via unique index

Jossec101:
Add partial unique indexes on WalletWithdrawalRequestPSBT and ChannelOperationRequestPSBT filtered on IsTemplatePSBT so concurrent GenerateTemplatePSBT calls can no longer persist two templates with different txids, which caused the approver to sign a transaction that PsbtApprovalValidator subsequently rejected.

Introduce GetSingleTemplatePsbt() helpers that throw on data-integrity violations instead of silently picking one, and preserve the convention FK indexes that would otherwise be dropped once an explicit index covers the FK column.

### DIFF
--- a/src/Data/Models/ChannelOperationRequest.cs
+++ b/src/Data/Models/ChannelOperationRequest.cs
@@ -161,6 +161,24 @@ namespace NodeGuard.Data.Models
         public int NumberOfSignaturesCollected => ChannelOperationRequestPsbts == null ? 0 : ChannelOperationRequestPsbts.Count(x => !x.IsFinalisedPSBT && !x.IsTemplatePSBT && !x.IsInternalWalletPSBT);
 
         /// <summary>
+        /// The single template PSBT of this request, or null when none has been generated yet.
+        /// The database enforces at most one (IX_ChannelOperationRequestPSBTs_Template); finding more is a
+        /// data-integrity fault and this throws rather than silently picking one.
+        /// </summary>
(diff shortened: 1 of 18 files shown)


src/Data/Models/ChannelOperationRequest.cs
```diff
@@ -160,6 +160,24 @@ public decimal Amount
         [NotMapped]
         public int NumberOfSignaturesCollected => ChannelOperationRequestPsbts == null ? 0 : ChannelOperationRequestPsbts.Count(x => !x.IsFinalisedPSBT && !x.IsTemplatePSBT && !x.IsInternalWalletPSBT);
 
+        /// <summary>
+        /// The single template PSBT of this request, or null when none has been generated yet.
+        /// The database enforces at most one (IX_ChannelOperationRequestPSBTs_Template); finding more is a
+        /// data-integrity fault and this throws rather than silently picking one.
```
markettes:
```suggestion
        /// data-integrity fault and this throws rather than silently picking one, because the template is
        /// what every approval is validated against and what the internal wallet co-signs.
```

markettes (dismissed):
There is quite duplicated code in these related classes and that makes them harder to maintain, we should think about abstracting some of this into `IBitcoinRequest`. Not saying that we should do it now, but we should add this to a future refactor for sure IMO.

markettes (approved):

